PR #114, opened 2023-01-11
migrate remaining unit tests for 2.0

rymorale:
## Types of changes

## Checklist:


code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/MessagingExtension.java
```diff
@@ -459,7 +454,7 @@ private static Map<String, Object> getXdmData(final String eventType, final Stri
         final Map<String, Object> trackingMap = new HashMap<>();
         final Map<String, Object> customActionMap = new HashMap<>();
 
-        if (actionId != null) {
+        if (!StringUtils.isNullOrEmpty(actionId)) {
```
sbenedicadb, line 457: is empty string valid?

sbenedicadb (approved):
looks good.  just one question about the change to disallow empty string which is a change in behavior potentially for reporting

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/MessagingExtension.java
rymorale: double checked the ios implementation and an empty string is valid. I'll revert this change.

rymorale: looking into this further, imo an empty string is not valid. the action id is extracted in this line:
```java
final String actionId = eventData.optString(MessagingConstants.EventDataKeys.Messaging.TRACK_INFO_KEY_ACTION_ID, null);
```
with the check above previously checking for null string only. i think updating it to using the StringUtils method is not necessary as the value will always be a valid string or null.

PR #149, opened 2023-03-03
provide default implementation for Message.getAutoTrack()

rymorale:
## Description
- Provide a default implementation for getAutoTrack() in the Message interface as it was previously present in the now internal InternalMessage class.
- also fix logging bug in MessagingExtension

## Types of changes

## Checklist:

### DIFF
--- a/code/messaging/src/main/java/com/adobe/marketing/mobile/Message.java
+++ b/code/messaging/src/main/java/com/adobe/marketing/mobile/Message.java
@@ -18,6 +18,8 @@ import android.webkit.WebView;
 import com.adobe.marketing.mobile.messaging.internal.WebViewJavascriptInterface;
 
 public interface Message {
+    boolean autoTrack = true;
+
     /**
      * Dispatch tracking information via a Messaging request content event.
      *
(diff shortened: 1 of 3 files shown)


code/messaging/src/main/java/com/adobe/marketing/mobile/Message.java
```diff
@@ -79,4 +81,13 @@ public interface Message {
...
      */
     void setAutoTrack(boolean enabled);
+
+    /**
+     * Gets the {@link Message}'s auto tracking preference.
+     *
+     * @return {@code boolean} containing the auto tracking preference.
+     */
+    default boolean getAutoTrack() {
+        return autoTrack;
```
praveek, line 91: Is true the default value? If so, can you remove the `autoTrack` variable and just return "true"

rymorale: makes sense as true is default.

code/messaging/src/main/java/com/adobe/marketing/mobile/Message.java
```diff
@@ -79,4 +79,13 @@ public interface Message {
      * @param enabled {@code boolean} if true, tracking is done automatically for {@code Message} show, dismiss, and triggered events.
      */
     void setAutoTrack(boolean enabled);
```
swarna04, line 81: Should this be `final` boolean enabled?

swarna04 (approved):

code/messaging/src/main/java/com/adobe/marketing/mobile/Message.java
rymorale: i'll update this

PR #161, opened 2023-03-22
Add FeedItem class + unit tests

rymorale:
## Types of changes

## Checklist:


code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/internal/FeedItem.java
```diff
@@ -0,0 +1,89 @@
...
+import org.json.JSONException;
+import org.json.JSONObject;
+
+import java.io.IOException;
+import java.io.ObjectInputStream;
+import java.io.ObjectOutputStream;
+import java.io.Serializable;
+import java.util.Map;
+
+class FeedItem implements Serializable {
```
swarna04, line 27: This needs to be a public class. Also, we do not need to cache (persist) this object and it can remain just in memory so there is no need to conform to Serializable.

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/internal/FeedItem.java
```diff
@@ -0,0 +1,89 @@
...
+    // Contains a URL to be opened if the user interacts with the feed item
+    String actionUrl;
+    // Required if actionUrl is provided. Text to be used in title of button or link in feed item
+    String actionTitle;
+    // Represents when this feed item went live. Represented in seconds since January 1, 1970
+    long publishedDate;
+    // Represents when this feed item expires. Represented in seconds since January 1, 1970
+    long expiryDate;
+    // Contains additional key-value pairs associated with this feed item
+    Map<String, Object> meta;
```
swarna04, line 43: Let's make the class members final and private. We can expose getters for accessing them.

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/internal/FeedItem.java
```diff
@@ -0,0 +1,89 @@
...
+        this.body = body;
+        this.imageUrl = imageUrl;
+        this.actionUrl = actionUrl;
+        this.actionTitle = actionTitle;
+        this.publishedDate = publishedDate;
+        this.expiryDate = expiryDate;
+        this.meta = meta;
+    }
+
+    static FeedItem create(final String title, final String body, final String imageUrl, final String actionUrl, final String actionTitle, final long publishedDate, final long expiryDate, final Map<String, Object> meta) {
```
swarna04, line 56: Should we use a builder here instead which would accept the required fields in the constructor and provide setters for the optional ones?

swarna04 (changes requested):

code/messaging/src/main/java/com/adobe/marketing/mobile/FeedItem.java
```diff
@@ -0,0 +1,152 @@
...
+
+        public Builder setMeta(final Map<String, Object> meta) {
+            throwIfAlreadyBuilt();
+
+            feedItem.meta = meta;
+            return this;
+        }
+
+        public FeedItem build() {
+            if (StringUtils.isNullOrEmpty(feedItem.title) || StringUtils.isNullOrEmpty(feedItem.body) || feedItem.publishedDate <= 0 || feedItem.expiryDate <= 0) {
```
swarna04, line 104: nit: I believe you do not need this check here during object construction. You can do this validation when you instantiate a FeedItem object using the provided event data to ensure required fields are present. You can also pinpoint the specific field that's missing as you parse it.

swarna04 (approved):

code/messaging/src/main/java/com/adobe/marketing/mobile/FeedItem.java
rymorale: the current behavior matches the iOS side which will return a nil if the FeedItem data is missing a required parameter when decoding. i'll keep it like this for now but can update in the future if needed.

PR #188, opened 2023-08-03
[MOB-187674] - Display push notification using AdobeMessagingServices

PravinPK:


code/messaging/src/main/java/com/adobe/marketing/mobile/AdobeMessagingService.java
```diff
@@ -0,0 +1,73 @@
...
+
+    @Override
+    public void onMessageReceived(final @NonNull RemoteMessage remoteMessage) {
+        super.onMessageReceived(remoteMessage);
+        handleRemoteMessage(this, remoteMessage);
+    }
+
+    public static boolean handleRemoteMessage(final @NonNull Context context, final @NonNull RemoteMessage remoteMessage) {
+        if (!isAdobePushNotification(remoteMessage)) {
+            Log.debug(MessagingPushConstants.LOG_TAG, SELF_TAG, "The received push message is not generated from Adobe Journey Optimizer, Messaging extension is ignoring to display the push notification.");
```
rymorale, line 43: how about 
```
; the Messaging extension will ignore the push notification.
```

code/messaging/src/main/java/com/adobe/marketing/mobile/MessagingPushBuilder.java
```diff
@@ -0,0 +1,196 @@
...
+        this(payload, context, new MessagingPushUtils(context));
+    }
+
+    /**
+     * Constructor for testing purposes.
+     * @param payload {@link MessagingPushPayload} the payload received from the push notification
+     * @param context the application {@link Context}
+     * @param utils {@link MessagingPushUtils} the utils class
+     */
+    MessagingPushBuilder(final @NonNull MessagingPushPayload payload, final @NonNull Context context, final @NonNull MessagingPushUtils utils) {
```
rymorale, line 57: add the @VisibleForTesting annotation here

code/messaging/src/main/java/com/adobe/marketing/mobile/MessagingPushBuilder.java
```diff
@@ -0,0 +1,196 @@
...
+    private String createChannelAndGetChannelID(final MessagingPushPayload payload, final Context context)  {
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.O) {
+            // For Android versions below O, no channel is created. Just return the obtained channel ID.
+            return payload.getChannelId() == null ? DEFAULT_CHANNEL_ID : payload.getChannelId();
+        } else {
+            // For Android versions O and above, create a channel if it does not exist and return the channel ID.
+            final NotificationManager notificationManager = (NotificationManager) context.getSystemService(Context.NOTIFICATION_SERVICE);
+            final String channelIdFromPayload = payload.getChannelId();
+
+            // if a channel from the payload
```
rymorale, line 108: looks like this comment is incomplete. can you clarify it?

code/messaging/src/main/java/com/adobe/marketing/mobile/MessagingPushBuilder.java
```diff
@@ -0,0 +1,196 @@
...
+            } else {
+                Log.warning(MessagingPushConstants.LOG_TAG, SELF_TAG, "No valid small icon found. Notification will not be displayed.");
+            }
+        }
+    }
+
+    /**
+     * Sets the sound for the notification.
+     * If a sound is received from the payload, the same is used.
+     * If a sound is not received from the payload, the default sound is used
```
rymorale, line 164: can we add info on the format of the sound uri used? e.g. is it just `filename` or `filename.ext`.

code/messaging/src/main/java/com/adobe/marketing/mobile/MessagingPushPayload.java
```diff
@@ -212,6 +213,25 @@ private int getNotificationPriorityFromString(final String priority) {
...
+                    return NotificationManager.IMPORTANCE_MIN;
+                case NotificationPriorities.PRIORITY_LOW:
+                    return NotificationManager.IMPORTANCE_LOW;
+                case NotificationPriorities.PRIORITY_HIGH:
+                    return NotificationManager.IMPORTANCE_HIGH;
+                case NotificationPriorities.PRIORITY_MAX:
+                    return NotificationManager.IMPORTANCE_MAX;
+                case NotificationPriorities.PRIORITY_DEFAULT:
+                default:
+                    return NotificationManager.IMPORTANCE_NONE;
```
rymorale, line 231: maybe we could use default importance instead? `NotificationManager.IMPORTANCE_DEFAULT`

code/messaging/src/main/java/com/adobe/marketing/mobile/MessagingPushUtils.java
```diff
@@ -0,0 +1,85 @@
...
+        this.context = context;
+    }
+
+    Bitmap download(final String url) {
+        Bitmap bitmap = null;
+        try {
+            final URL imageUrl = new URL(url);
+            final HttpURLConnection connection = (HttpURLConnection) imageUrl.openConnection();
+            bitmap = BitmapFactory.decodeStream(connection.getInputStream());
+            connection.disconnect();
```
rymorale, line 42: could the networking service be leveraged here rather than writing new networking code? or would it be overkill for this purpose?

rymorale (changes requested):
looks good with just a few minor things to resolve

code/messaging/src/main/java/com/adobe/marketing/mobile/AdobeMessagingService.java
```diff
@@ -0,0 +1,73 @@
...
+        return true;
+    }
+
+
+    // This method looks at the remote message and determines if it is an Adobe push notification
+    // by checking if it contains the title and body keys.
+    // This is a temporary solution until we have a better way to identify Adobe push notifications.
+    // @param remoteMessage the remote message to check
+    // @return true if the remote message is an Adobe push notification, false otherwise.
+    private static boolean isAdobePushNotification(final @NonNull RemoteMessage remoteMessage) {
```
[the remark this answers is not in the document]
PravinPK: Good question. We are planning to introduce a new key in the payload whose purpose is to identify the source of push notifications. 

The current implementation will identity
 Assurance and AJO - regular push.    (it contains for the mandatory title key adb_title)
 AJO -  silent push (it contains the tracking key _xdm)
Assurance  - silent push. (Will be handled at this moment)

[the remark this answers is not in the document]
PravinPK: I have added a todo to reimplement when the services are updated

code/messaging/src/main/java/com/adobe/marketing/mobile/MessagingPushBuilder.java
```diff
@@ -0,0 +1,196 @@
...
+        } else {
+            // For Android versions O and above, create a channel if it does not exist and return the channel ID.
+            final NotificationManager notificationManager = (NotificationManager) context.getSystemService(Context.NOTIFICATION_SERVICE);
+            final String channelIdFromPayload = payload.getChannelId();
+
+            // if a channel from the payload
+            if (channelIdFromPayload != null && notificationManager.getNotificationChannel(channelIdFromPayload) != null) {
+                Log.debug(MessagingPushConstants.LOG_TAG, SELF_TAG, "Channel exists for channel ID: " + channelIdFromPayload + ". Using the same for push notification.");
+                return channelIdFromPayload;
+            } else {
```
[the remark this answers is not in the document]
PravinPK: I think it's a good idea to fall back to default channel ID if the channel does not exist.
Creating channels for every channel ID coming from the notification, can pollute the apps notification settings page.

IMO the only app's developer should have the control and assign all the required properties for the channel

code/messaging/src/main/java/com/adobe/marketing/mobile/MessagingPushPayload.java
PravinPK: that sounds right. Thanks for the explanation

code/messaging/src/main/java/com/adobe/marketing/mobile/MessagingPushUtils.java
PravinPK: I am thinking the same too. I dont think there is any negative impact using the HttpURLConnection as well. It might be an overkill.  Moreover, we only have asyn call on network service and that will switch the creating on notification to  network thread.

code/messaging/src/main/java/com/adobe/marketing/mobile/MessagingPushUtils.java
```diff
@@ -0,0 +1,85 @@
...
+import androidx.annotation.NonNull;
+import com.adobe.marketing.mobile.services.Log;
+import com.adobe.marketing.mobile.util.StringUtils;
+import java.io.IOException;
+import java.net.HttpURLConnection;
+import java.net.URL;
+
+class MessagingPushUtils {
+    private static final String SELF_TAG = "PushNotificationMediaDownloader";
+    private final Context context;
```
prudrabhat, line 30: Recommend using the `context` as a parmeter to the utility function where required instead of holding a strong reference to Context here. The caller can fetch it via `ServiceProvider.getInstance.getApplication().getApplicationContex()`.  That will allow you to make this a true singleton and the methods static.

code/messaging/src/main/java/com/adobe/marketing/mobile/MessagingPushBuilder.java
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    /**
+     * Sets the large icon for the notification.
+     * If a large icon url is received from the payload, the image is downloaded and the notification style is set to BigPictureStyle.
+     * If large icon url is not received from the payload, default style is used for the notification.
+     * @param notificationBuilder the notification builder
+     */
+    private void setLargeIcon(final NotificationCompat.Builder notificationBuilder) {
+        if (!StringUtils.isNullOrEmpty(payload.getImageUrl())) {
```
prudrabhat, line 184: nit: Eliminate the arrow anti-pattern by returning early.

```
if (StringUtils.isNullOrEmpty(payload.getImageUrl())) return;
...
...
if (bitmap == null) return;

....
```

code/messaging/src/main/java/com/adobe/marketing/mobile/MessagingPushBuilder.java
```diff
@@ -0,0 +1,199 @@
...
+                NotificationCompat.BigPictureStyle bigPictureStyle = new NotificationCompat.BigPictureStyle();
+                bigPictureStyle.bigLargeIcon(bitmap);
+                notificationBuilder.setStyle(bigPictureStyle);
+            }
+
+        }
+    }
+
+    private boolean isValidIcon(final int icon) {
+        return icon > 0 ? true : false;
```
prudrabhat, line 197: can be replaced with `return icon > 0`

code/messaging/src/main/java/com/adobe/marketing/mobile/MessagingPushBuilder.java
```diff
@@ -0,0 +1,199 @@
...
+     * If a sound is received from the payload, the same is used.
+     * If a sound is not received from the payload, the default sound is used
+     * The sound name from the payload should also include the format of the sound file. eg: sound.mp3
+     * @param notificationBuilder the notification builder
+     */
+    private void setSound(final NotificationCompat.Builder notificationBuilder) {
+        if(!StringUtils.isNullOrEmpty(payload.getSound())) {
+            notificationBuilder.setSound(utils.getSoundUriForResourceName(payload.getSound()));
+        }
+        notificationBuilder.setDefaults(Notification.DEFAULT_ALL);
```
prudrabhat, line 174: Looks like this API is deprecated in level 26. Does using this `notificationBuilder.setSound(RingtoneManager.getDefaultUri(RingtoneManager.TYPE_NOTIFICATION));` result in the same behavior?

code/messaging/src/main/java/com/adobe/marketing/mobile/MessagingPushUtils.java
```diff
@@ -0,0 +1,85 @@
...
+import android.net.Uri;
+import androidx.annotation.NonNull;
+import com.adobe.marketing.mobile.services.Log;
+import com.adobe.marketing.mobile.util.StringUtils;
+import java.io.IOException;
+import java.net.HttpURLConnection;
+import java.net.URL;
+
+class MessagingPushUtils {
+    private static final String SELF_TAG = "PushNotificationMediaDownloader";
```
prudrabhat, line 29: Did you intentionally use tag value?

PravinPK: Nope this was my old class name. Will need to change them. Good catch

code/messaging/src/main/java/com/adobe/marketing/mobile/MessagingPushBuilder.java
PravinPK: Yeap. `setDefaults` is deprecated. From API 26, sounds can only be set on the notification channel.
they have also deprecated the setSound API https://developer.android.com/reference/android/app/Notification.Builder#setSound(android.net.Uri)

PravinPK: Let's keep.
notificationBuilder.setSound(RingtoneManager.getDefaultUri(RingtoneManager.TYPE_NOTIFICATION)); for setting sound for API 26 and below and make sure we document that fact that sound has to set on the Notification channel

rymorale (approved):

code/messaging/src/main/java/com/adobe/marketing/mobile/MessagingPushBuilder.java
```diff
@@ -0,0 +1,201 @@
...
+                       final Context context){
+        final String channelId = createChannelAndGetChannelID(payload, context);
+
+        // Create the notification
+        final NotificationCompat.Builder builder = new NotificationCompat.Builder(context, channelId);
+        builder.setContentTitle(payload.getTitle());
+        builder.setContentText(payload.getBody());
+        builder.setNumber(payload.getBadgeCount());
+        builder.setPriority(payload.getNotificationPriority());
+        builder.setAutoCancel(true);
```
prudrabhat, line 60: nit: why not chain these calls :)

code/messaging/src/main/java/com/adobe/marketing/mobile/MessagingPushUtils.java
```diff
@@ -0,0 +1,88 @@
...
+    static Bitmap download(final String url) {
+        Bitmap bitmap = null;
+        try {
+            final URL imageUrl = new URL(url);
+            final HttpURLConnection connection = (HttpURLConnection) imageUrl.openConnection();
+            bitmap = BitmapFactory.decodeStream(connection.getInputStream());
+            connection.disconnect();
+        } catch(IOException e) {
+            Log.warning(MessagingPushConstants.LOG_TAG, SELF_TAG, "Failed to download push notification large image from url: %s", url);
+        }
```
prudrabhat, line 46: Can you also close the connection and the inputstream in a `finally` block.

prudrabhat (approved):
Couple minor comments.

PR #220, opened 2024-01-12
Jetpack compose changes for messaging

spoorthipujariadobe:
## Description

1. Updated `minSDKVersion` from 19 -> 21 to align with Core
2. Updated Kotlin `languageVersion` and `apiVersion` from 1.4 -> 1.5 as 1.4 will be depcreated
3. Removed deprecated `kotlin-android-extensions` and replaced with recommended view bindings
4. Removed the deprecated `registerExtension` API
5. Moved from using `ServiceProvider.getInstance().getUIService().showUrl` -> `ServiceProvider.getInstance().getUriService().openUri`
6. Moved from using `uiService.createFullscreenMessage` -> `uiService.create`,`FullscreenMessageDelegate` ->  `InAppMessageEventListener` ,`MessageSettings` -> `InAppMessageSettings`, `MessagingDelegate` -> `PresentationDelegate`
7. Created new public `PresentableMessageMapper` to expose `Message` class behind a `Presentable` that can be used by custom delegate implementers

## Types of changes

## Checklist:


code/build.gradle
```diff
@@ -2,7 +2,7 @@
 
 buildscript {
     ext {
-        kotlin_version = '1.4.0'
+        kotlin_version = '1.7.20'
```
spoorthipujariadobe, line 5: Note: These versions will be updated to be used from Gradle Plugin in subsequent PR for 3.x migration

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/PresentableMessageMapper.java
```diff
@@ -0,0 +1,295 @@
...
+import com.adobe.marketing.mobile.util.DataReader;
+import com.adobe.marketing.mobile.util.DefaultPresentationUtilityProvider;
+import com.adobe.marketing.mobile.util.MapUtils;
+import com.adobe.marketing.mobile.util.StringUtils;
+
+import java.lang.ref.WeakReference;
+import java.util.HashMap;
+import java.util.Map;
+
+public final class PresentableMessageMapper {
```
prudrabhat, line 38: Discussed with Spoorthi on a call about modifying this class to prevent exposing the mapping logic. 

The reason a mapping is being held between the Presentable `id` and the `Message` is because customers who implement a `PresentationDelegate` can chose to manage the display and dismissal in `PresentationDelegate.canShow()` and also use it to enable and disable tracking for the Message associated with the Presentable.  The app implementer does not necessarily need to know about the mapping.
Instead of exposing the `PresentableMessageMapper` and the underlying mapping employed, we agreed that a simple utility method that provides a `Message` given a `Presentable<InAppMessage>` is sufficient for this use case.
 
```
class MessageUtilities { // or some appropriate name
     ....
     fun getMessageFromPresentable(@NonNull Presentable<InAppMessage> presentable): Message? {
        val id = getPresentation().getId()
        return PresentableMapper.getMessageFromPresentableId(id)
     }
}
```

There is a need to ensure that the map is cleaned up, but since the `PresentationDelegate` implementer can hold references to the `Presentable<InAppMessage>` and can invoke `show()`/`dismiss()` later (and likely need the id mapping) we cannot cleanup on show or on error. Given that the accumulation of several presentables is not common (perhaps rare also), we plan to check for the references that went out of context while creating new InAppMessages in the extension.

This allows us to make PresentableMessageMapper private to the extension and limit exposing the internal logic.

spoorthipujariadobe: Changes as per discussion

sbenedicadb (approved):
looks good - let's make sure we're keeping track of 2.x -> 3.x implementation changes required so we can update our docs accordingly at the end 👍

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/PresentableMessageMapper.java
```diff
@@ -0,0 +1,325 @@
...
+            aepMessage = uiService.create(new InAppMessage(settings, new MessagingFullscreenEventListener()), new DefaultPresentationUtilityProvider());
+        }
+
+        /**
+         * Dispatch tracking information via a Messaging request content event.
+         *
+         * @param interaction {@code String} containing the interaction which occurred
+         * @param eventType   {@link MessagingEdgeEventType} enum containing the Event Type to be used for the ensuing Edge Event
+         */
+        public void track(final String interaction, final MessagingEdgeEventType eventType) {
```
prudrabhat, line 200: Looks like this is an overridden method. Can you add the `@Override` annotation for clarity. Same for below methods overridden from `Message`.

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/PresentableMessageMapper.java
```diff
@@ -0,0 +1,325 @@
...
+        private InAppMessageSettings InAppMessageSettingsFromMap(final Map<String, Object> rawSettings,
+                                                                 final String content,
+                                                                 final Map<String, String> assetMap) {
+            int width, height, verticalInset, horizontalInset;
+            String backdropColor;
+            float backdropOpacity;
+            float cornerRadius;
+            boolean uiTakeover;
+            InAppMessageSettings.MessageAlignment verticalAlign, horizontalAlign;
+            InAppMessageSettings.MessageAnimation displayAnimation, dismissAnimation;
```
prudrabhat, line 284: Can you combine these with initialization?

```
int width = DataReader.optInt(rawSettings, MessagingConstants.EventDataKeys.MobileParametersKeys.WIDTH, FILL_SCREEN);
...
```

code/messaging/src/phone/java/com/adobe/marketing/mobile/messaging/MessagingUtils.java
```diff
@@ -123,4 +128,21 @@ public static Surface scopeToSurface(final String scope) {
...
+    // Presentable message utils
+    // ========================================================================================
+    /**
+     * Returns the {@code Message} object from the provided {@code Presentable<InAppMessage>} object.
+     *
+     * @param presentable A {@link Presentable<InAppMessage>} object
+     * @return the {@link Message} object linked to the presentable or null if the presentable is null
+     */
+    @Nullable
+    public static Message getMessageForInAppMessagePresentable(final Presentable<InAppMessage> presentable) {
```
prudrabhat, line 142: I feel that this method can be renamed to `getMessageForPresentable` given that the signature explicitly needs an InAppMessage presentable.

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/PresentableMessageMapper.java
```diff
@@ -0,0 +1,325 @@
...
+
+        // ui management
+        public void show() {
+            if (aepMessage != null) {
+                aepMessage.show();
+            }
+        }
+
+        public void dismiss(final boolean suppressAutoTrack) {
+            if (aepMessage != null) {
```
prudrabhat, line 216: Can you git rid of the nested if here by returning early.

```
if (aepMessage == null) return;

if (autoTrack && !suppressAutoTrack) {
       track(null, MessagingEdgeEventType.IN_APP_DISMISS);
 }
aepMessage.dismiss();
}

```

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/PresentableMessageMapper.java
```diff
@@ -0,0 +1,325 @@
...
+                if (autoTrack && !suppressAutoTrack) {
+                    track(null, MessagingEdgeEventType.IN_APP_DISMISS);
+                }
+
+                aepMessage.dismiss();
+            }
+        }
+
+        void trigger() {
+            if (aepMessage != null) {
```
prudrabhat, line 226: nit: return early if there is no message to simplify if arrow pattern.

prudrabhat (approved):
👍🏼

PR #291, opened 2024-10-22
Add Messaging UI event observer and ContentCardMapper

rymorale:
## Description
- Add MessagingUIEventObserver to handle content card tracking.
- Add a ContentCardMapper singleton that will be used to store valid content card schema data objects. The schema data can be retrieved later on for tracking purposes. The content card schema data object map is updated when content card rules are evaluated and when content cards are disqualified.

## Types of changes

## Checklist:


code/messaging/src/main/java/com/adobe/marketing/mobile/aepuitemplates/SmallImageTemplate.kt
```diff
@@ -28,7 +29,7 @@ import com.adobe.marketing.mobile.aepuitemplates.utils.AepUITemplateType
  * @property dismissBtn The details for the small image template dismiss button.
  */
 data class SmallImageTemplate(
-    val id: String,
+    val propositionId: String,
```
prudrabhat, line 32: We should let this remain an `id` so as to make it generic and not closely coupled with Messaging.

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardMapper.kt
```diff
@@ -0,0 +1,50 @@
...
+
+        @Volatile
+        private var instance: ContentCardMapper? = null
+
+        @JvmStatic
+        fun getInstance() =
+            instance ?: synchronized(this) {
+                instance ?: ContentCardMapper().also { instance = it }
+            }
+    }
```
prudrabhat, line 31: Can we change this to a Bill Pugh singleton.

```
class Singleton private constructor() {
    companion object {
        val instance: Singleton by lazy { Holder.INSTANCE }
    }

    private object Holder {
        val INSTANCE = Singleton()
    }
}
```

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/MessagingUIEventObserver.kt
```diff
@@ -0,0 +1,57 @@
...
+            is UIEvent.Display -> {
+                Log.trace(MessagingConstants.LOG_TAG, SELF_TAG, "SmallImageUI Displayed")
+                handleSmallImageTracking(event.aepUi.getTemplate(), null, MessagingEdgeEventType.DISPLAY)
+            }
+            is UIEvent.Interact -> {
+                if (callback?.onCardClick(event.aepUi.getTemplate()) == true) {
+                    Log.trace(MessagingConstants.LOG_TAG, SELF_TAG, "SmallImageUI Click")
+                    handleSmallImageTracking(event.aepUi.getTemplate(), "clicked", MessagingEdgeEventType.INTERACT)
+                }
+            }
```
prudrabhat, line 37: The `Interact` event's action should be more elaborate than a String. So that we can have more properties Can you make it a set of sealed `Actions` with `Click` being one of the actions? We should also be able to provide a way pass the url/uri or something wrt to the actions. Note that these actions will be different for different interactions (swipe, long press etc)

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/MessagingUIEventObserver.kt
```diff
@@ -0,0 +1,57 @@
...
+class MessagingUiEventObserver(private val callback: ContentCardCallback?) : AepUIEventObserver {
+    companion object {
+        private const val SELF_TAG = "MessagingUiEventObserver"
+    }
+
+    override fun onEvent(event: UIEvent<*, *>) {
+        when (event) {
+            is UIEvent.Display -> {
+                Log.trace(MessagingConstants.LOG_TAG, SELF_TAG, "SmallImageUI Displayed")
+                handleSmallImageTracking(event.aepUi.getTemplate(), null, MessagingEdgeEventType.DISPLAY)
```
prudrabhat, line 30: I understand that this is only implemented for SmalImageTemplate, however can we make it easier for forth coming templates by adding a filter based on template type and then handle the small image events.

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/MessagingUIEventObserver.kt
```diff
@@ -0,0 +1,57 @@
...
+
+    private fun handleSmallImageTracking(template: AepUITemplate, interaction: String?, eventType: MessagingEdgeEventType) {
+        val template = template as SmallImageTemplate
+        val contentCardSchemaData = ContentCardMapper.getInstance().getContentCardSchemaDataForPropositionId(template.propositionId)
+        contentCardSchemaData?.track(interaction, eventType)
+    }
+}
+
+interface ContentCardCallback {
+    fun onCardClick(template: AepUITemplate): Boolean
```
prudrabhat, line 55: This seems ambiguous - does this mean that the card has been clicked or that the button has been clicked? 
We should also have a way to report and distinguish when multiple buttons have been clicked.

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/MessagingUIEventObserver.kt
```diff
@@ -0,0 +1,57 @@
...
+package com.adobe.marketing.mobile.messaging
+
+import com.adobe.marketing.mobile.MessagingEdgeEventType
+import com.adobe.marketing.mobile.aepcomposeui.interactions.UIEvent
+import com.adobe.marketing.mobile.aepcomposeui.observers.AepUIEventObserver
+import com.adobe.marketing.mobile.aepuitemplates.AepUITemplate
+import com.adobe.marketing.mobile.aepuitemplates.SmallImageTemplate
+import com.adobe.marketing.mobile.services.Log
+
+class MessagingUiEventObserver(private val callback: ContentCardCallback?) : AepUIEventObserver {
```
prudrabhat, line 21: Suggest renaming this to `ContentCardEventObserver()`

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/MessagingUIEventObserver.kt
prudrabhat, line 55: Can you elaborate on what the return value means here?

prudrabhat (changes requested):

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/MessagingUIEventObserver.kt
rymorale: i'll add kdoc for the interface but i was thinking it would be used to signal if we should do tracking for the interaction or if it has been handled manually.

rymorale: i'll expand the actions captured by the callback

code/messaging/src/main/java/com/adobe/marketing/mobile/aepcomposeui/interactions/UIEvent.kt
```diff
@@ -61,8 +61,8 @@ sealed interface UIEvent<T : AepUITemplate, S : AepCardUIState> {
      *
      * @param T represents UI template model associated like [SmallImageTemplate], which backs the composable on which the event has occurred.
```
spoorthipujariadobe, line 62: Wondering if we can change Interact to have these fields

1. aepUi
2. action of type `UIAction` which would have different interact enums like `CLICK`, `SWIPE` etc.
3. id: Might be useful to look up in schema data? not too sure
4. actionUrl of type String. Since every clickable element's model (eg: AepButton) has actionUrl guessing it should be easy to add here

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardRulesEngine.java
```diff
@@ -72,6 +72,9 @@ Map<Surface, List<PropositionItem>> evaluate(@NonNull final Event event) {
                 continue;
             }
 
+            // store valid content card schema data in the ContentCardMapper for later use
+            ContentCardMapper.getInstance().storeContentCardSchemaData(propositionAsContentCard);
```
spoorthipujariadobe, line 76: Wondering if we should add this logic in `addOrReplaceContentCards` where the qualification happens. We can maintain a smaller mapping for qualified cards only instead of all cards returned by server. Thoughts?

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/MessagingUIEventObserver.kt
spoorthipujariadobe: How about these fields for Interact?

1. aepUi
2. action of type UIAction which would have different interact enums like CLICK, SWIPE etc.
3. id: Might be useful to look up in meta data? not too sure
4. actionUrl of type String. Since every clickable element's model (eg: AepButton) has actionUrl guessing it should be easy to add here

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/MessagingUIEventObserver.kt
```diff
@@ -0,0 +1,57 @@
...
+            is UIEvent.Interact -> {
+                if (callback?.onCardClick(event.aepUi.getTemplate()) == true) {
+                    Log.trace(MessagingConstants.LOG_TAG, SELF_TAG, "SmallImageUI Click")
+                    handleSmallImageTracking(event.aepUi.getTemplate(), "clicked", MessagingEdgeEventType.INTERACT)
+                }
+            }
+            is UIEvent.Dismiss -> {
+                if (callback?.onCardDismiss(event.aepUi.getTemplate()) == true) {
+                    Log.trace(MessagingConstants.LOG_TAG, SELF_TAG, "SmallImageUI Dismissed")
+                    handleSmallImageTracking(event.aepUi.getTemplate(), null, MessagingEdgeEventType.DISMISS)
```
spoorthipujariadobe, line 41: Do we also need to send the disqualify event here?

spoorthipujariadobe: Oh nevermind it automatically happens based on event type

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardRulesEngine.java
rymorale: i think this is a good idea. i can make this change.

code/messaging/src/main/java/com/adobe/marketing/mobile/aepcomposeui/utils/UIAction.kt
```diff
@@ -11,9 +11,15 @@
...
 
-class UIAction {
-    companion object {
-        const val CLICK = "click"
-        const val EXPAND = "expand"
+enum class UIAction(private val action: String) {
+    CLICK("click"),
+    DOUBLE_CLICK("doubleClick"),
+    DRAG("drag"),
+    EXPAND("expand"),
```
prudrabhat, line 18: - `Expand` is a special type of click (similar to dismiss) and the result of the click is an expand. IMO this should not be a separate action. 
- Let us only add actions that we need for now (only clicks I think for now)
- If this is also made a sealed class, we can let each action have its own set of properties while allowing consumers to use `when` to identify the action.

sealed class UiAction(id: String) {
    data class Click(val id: String, val actionUrl: String) : UiEvent(id)  // Represents a click with some action
   ....
}

code/messaging/src/main/java/com/adobe/marketing/mobile/aepcomposeui/interactions/UIEvent.kt
```diff
@@ -27,42 +28,44 @@ sealed interface UIEvent<T : AepUITemplate, S : AepCardUIState> {
...
      * Example:
      * ```
-     * observer?.onEvent(AepUiEvent.Interact(ui, action = UIAction.CLICK))
+     * observer?.onEvent(AepUiEvent.Interact(ui, interactId = "purchaseID", action = UIAction.CLICK), actionUrl = "https://www.adobe.com")
      * ```
      */
     data class Interact<T : AepUITemplate, S : AepCardUIState>(
         val aepUi: AepUI<T, S>,
-        val action: String
+        val interactId: String?,
```
spoorthipujariadobe, line 57: nit: Can we rename this? This is the id of the field which was interacted with. Something like `interactedId`. Or maybe we can leave this out for now if its confusing

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardEventObserver.kt
```diff
@@ -0,0 +1,184 @@
...
+     */
+    fun onButtonClick(event: UIEvent.Interact<*, *>): Boolean
+
+    /**
+     * Callback to invoke when a content card dismiss button is clicked.
+     *
+     * @param event The [UIEvent.Interact] which occurred.
+     * @return true if tracking should be handled by the Messaging extension, false otherwise.
+     */
+    fun onDismissButtonClick(event: UIEvent.Interact<*, *>): Boolean
```
spoorthipujariadobe, line 183: Wondering if we should club all interactions to a generic `onInteract` since buttons is not mandatory. And it would consistent with iOS. Thoughts?

spoorthipujariadobe: Also do you think we need both `onDismissButtonClick` and `onCardDismiss`? Since there's no way customer can do the disqualification, can we just inform them about the dismiss so they can update the UI?

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardEventObserver.kt
```diff
@@ -0,0 +1,184 @@
...
+import com.adobe.marketing.mobile.services.Log
+
+class ContentCardEventObserver(private var callback: ContentCardCallback?) : AepUIEventObserver {
+    companion object {
+        private const val SELF_TAG = "ContentCardEventObserver"
+    }
+
+    override fun onEvent(event: UIEvent<*, *>) {
+        if (callback == null) {
+            callback = DefaultContentCardCallback()
```
spoorthipujariadobe, line 28: Why do we need the `DefaultContentCardCallback`?

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardEventObserver.kt
```diff
@@ -0,0 +1,184 @@
...
+        // content card buttons have an interact id, dismiss buttons will have the style as the interact id
+        if (!event.interactId.isNullOrEmpty()) {
+            val contentCardSchemaData = ContentCardMapper.instance.getContentCardSchemaDataForPropositionId(smallImageTemplate.id)
+            when (event.interactId) {
+                "none", "simple", "circle" -> { // handle content card dismiss button click
+                    if (callback?.onDismissButtonClick(event) == true) {
+                        Log.trace(MessagingConstants.LOG_TAG, SELF_TAG, "SmallImageUI dismiss button click")
+                        contentCardSchemaData?.track(event.action.toString(), MessagingEdgeEventType.INTERACT)
+                    }
+                }
```
spoorthipujariadobe, line 94: Wouldn't interaction with dismiss button produce `UIEvent.Dismiss` instead of `UIEvent.Click`?

code/messaging/src/main/java/com/adobe/marketing/mobile/aepcomposeui/interactions/UIEvent.kt
rymorale: this id is removed. the id will now be taken from the UIAction which occurred:
```kotlin
/**
 * Represents an action that can be performed on a UI component.
 */
sealed class UIAction {
    /**
     * UIAction to represent a click action that can be performed on a UI component
     * @property id unique identifier of the UI component
     * @property actionUrl optional URL to be opened when the UI component is clicked
     */
    data class Click(val id: String, val actionUrl: String?) : UIAction()
}
```

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardEventObserver.kt
```diff
@@ -0,0 +1,184 @@
...
+ */
+interface ContentCardCallback {
+
+    /**
+     * Callback to invoke when a content card is displayed.
+     *
+     * @param event The [UIEvent.Display] which occurred.
+     * @return true if tracking should be handled by the Messaging extension, false otherwise.
+     */
+    fun onCardDisplayed(event: UIEvent.Display<*, *>): Boolean
```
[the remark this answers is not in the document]
rymorale: i've removed the boolean return value and now more closely match the iOS implementation

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardEventObserver.kt
rymorale: this is also resolved as i've updated the SmallImageTemplateEventHandler to handle a generic `onInteractEvent`

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardEventObserver.kt
rymorale: I've removed this as there are no longer boolean values being returned

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardEventObserver.kt
rymorale: you are right, fixed this

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardEventObserver.kt
```diff
@@ -0,0 +1,77 @@
...
+    companion object {
+        private const val SELF_TAG = "ContentCardEventObserver"
+    }
+
+    override fun onEvent(event: UIEvent<*, *>) {
+        when (event) {
+            is UIEvent.Display -> {
+                when (val template = event.aepUi.getTemplate()) {
+                    is SmallImageTemplate -> {
+                        val eventHandler = SmallImageTemplateEventHandler(callback)
```
spoorthipujariadobe, line 29: nit: Do we need to create a new instance of `SmallImageTemplateEventHandler` for each event? Do you think we can create one when the `ContentCardEventObserver` is initialized?

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardEventObserver.kt
```diff
@@ -0,0 +1,77 @@
...
+    fun onDismiss()
+
+    /**
+     * Callback to invoke when a content card is interacted with.
+     *
+     * @param aepUI The AepUI instance that was interacted with.
+     * @param interactionId An optional string identifier for the interaction event.
+     * @param actionUrl An optional URL associated with the interaction.
+     */
+    fun onInteract(aepUI: AepUI<*, *>, interactionId: String?, actionUrl: String?)
```
spoorthipujariadobe, line 76: We might need the boolean return on only this one. Sorry, my bad 😶 
https://github.com/adobe/aepsdk-ui-ios/blob/v5.1.0-beta/Frameworks/AEPSwiftUI/Sources/ContentCards/ContentCardUIEventListening.swift#L40

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardEventObserver.kt
rymorale: the aepUI parameter is only available for the Display, Interact, and Dismiss events. i could cast the aepUI parameter outside of the `when` block but i would still need to determine the type before casting.

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardEventObserver.kt
rymorale: no prob, i'll get it added

code/messaging/src/main/java/com/adobe/marketing/mobile/aepcomposeui/components/SmallImageCard.kt
```diff
@@ -54,7 +54,7 @@ fun SmallImageCard(
     Card(
         modifier = Modifier
             .clickable {
-                observer?.onEvent(UIEvent.Interact(ui, UIAction.CLICK))
+                observer?.onEvent(UIEvent.Interact(ui, UIAction.Click(null, ui.getTemplate().actionUrl)))
```
spoorthipujariadobe, line 57: nit: `SmallImageTemplate` has an id (same as propositionId). Do you think it would be useful to pass that here?

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardEventObserver.kt
spoorthipujariadobe: Ah ok, understood. Do you think we can create the event handler for all types at class initialization and use the appropriate one here after the check? Something like
```

private lateinit val smallImageEventHandler
init {
    SmallImageTemplateEventHandler(callback)
    LargeImageTemplateEventHandler(callback)
    .....
}


override fun onEvent(event: UIEvent<*, *>) {
        when (event) {
            is UIEvent.Display -> {
                when (val template = event.aepUi.getTemplate()) {
                   is SmallImageTemplate -> {
                        val eventHandler = smallImageEventHandler.onEvent(event, template.id)
                    }
                }
            }
       }
}```

File: code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/MessagingEventHandler.kt [hunk 19]
```diff
@@ -0,0 +1,44 @@
...
+        private const val SELF_TAG = "MessagingEventHandler"
+    }
+
+    internal fun onEvent(event: UIEvent<*, *>, propositionId: String) {
+        var eventType: MessagingEdgeEventType? = null
+        if (event is UIEvent.Display) {
+            Log.trace(MessagingConstants.LOG_TAG, SELF_TAG, "${event.aepUi.getTemplate().getType()} Displayed")
+            callback?.onDisplay()
+            eventType = MessagingEdgeEventType.DISPLAY
+        } else if (event is UIEvent.Dismiss) {
```
spoorthipujariadobe, line 35: nit: Since we already have this `is UIEvent.Dismiss` check in `ContentCardEventObserver`, do you think it can be avoided by having two methods: `onDisplayEvent(UIEvent.Display)` and `onDismissEvent(UIEvent.Dismiss)`?

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/SmallImageTemplateEventHandler.kt
```diff
@@ -0,0 +1,78 @@
...
+                    Log.trace(
+                        MessagingConstants.LOG_TAG,
+                        SELF_TAG,
+                        "SmallImageUI opening URL: ${event.action.actionUrl}"
+                    )
+                    ServiceProvider.getInstance().uriService.openUri(event.action.actionUrl)
+                }
+
+                when (event.action.id) {
+                    "none", "simple", "circle" -> { // handle content card dismiss button click
```
spoorthipujariadobe, line 51: I ended up sending a `UIEvent.Dismiss` event on click of the dismiss button, instead of UIEvent.Click. Let me know what you think
https://github.com/adobe/aepsdk-messaging-android/pull/293/files#diff-f531a1c6d529c3b96a6b40d63609791b1ebb50b5a8334773633e9fd44db913c4R80

code/messaging/src/main/java/com/adobe/marketing/mobile/aepcomposeui/components/SmallImageCard.kt
rymorale: we can pass it here for ease of implementation

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/SmallImageTemplateEventHandler.kt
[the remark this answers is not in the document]
spoorthipujariadobe: This might not be needed since `SmallImageCard` dispatches `UIEvent.Dismiss` on dismiss button click
https://github.com/adobe/aepsdk-messaging-android/pull/293/files#diff-f531a1c6d529c3b96a6b40d63609791b1ebb50b5a8334773633e9fd44db913c4R80

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/MessagingEventHandler.kt
rymorale: this is now refactored with the changes that prashanth proposed.

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/SmallImageTemplateEventHandler.kt
rymorale: this makes sense, i'll update this

code/messaging/src/main/java/com/adobe/marketing/mobile/aepcomposeui/utils/UIAction.kt
```diff
@@ -11,9 +11,14 @@
...
 
-class UIAction {
-    companion object {
-        const val CLICK = "click"
-        const val EXPAND = "expand"
-    }
+/**
+ * Represents an action that can be performed on a UI component.
+ */
+sealed class UIAction {
```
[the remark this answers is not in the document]
rymorale: did you mean `UIAction` from utils?

code/messaging/src/main/java/com/adobe/marketing/mobile/aepcomposeui/utils/UIAction.kt
```diff
@@ -11,9 +11,14 @@
...
+    /**
+     * UIAction to represent a click action that can be performed on a UI component
+     * @property id optional unique identifier of the UI component
+     * @property actionUrl optional URL to be opened when the UI component is clicked
+     */
+    data class Click(val id: String?, val actionUrl: String?) : UIAction()
```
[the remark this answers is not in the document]
rymorale: there is no id in the case of the content card or dismiss button are being clicked. currently only buttons have `interactId`'s assigned. @spoorthipujariadobe, please correct me if i'm wrong.

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardEventObserver.kt
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * Callback to invoke when a content card is displayed.
+     */
+    fun onDisplay()
+
+    /**
+     * Callback to invoke when a content card is dismissed.
+     */
+    fun onDismiss()
```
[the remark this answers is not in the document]
rymorale: i've updated this to match the iOS implementation

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/SmallImageTemplateEventHandler.kt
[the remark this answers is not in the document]
rymorale: i've cleaned this up based on spoorthi's update

code/messaging/src/main/java/com/adobe/marketing/mobile/aepcomposeui/components/SmallImageCard.kt
```diff
@@ -49,7 +49,7 @@ fun SmallImageCard(
     AepCardComposable(
         cardStyle = style.cardStyle,
         onClick = {
-            observer?.onEvent(UIEvent.Interact(ui, UIAction.CLICK))
+            observer?.onEvent(UIEvent.Interact(ui, UIAction.Click(null, ui.getTemplate().actionUrl), ui.getTemplate().id))
```
spoorthipujariadobe, line 52: Are we missing passing `aepUi` here?

spoorthipujariadobe (approved):
One small question about ids in interact event, looks good otherwise

code/messaging/src/main/java/com/adobe/marketing/mobile/aepcomposeui/utils/UIAction.kt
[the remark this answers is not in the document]
spoorthipujariadobe: The content card id is probably the same as the template id. But dismiss button does not have one today. Please feel free to add it in `AepDismissButton` if you think it makes sense. Messaging can pass in some value like "dismiss" for it in the absence of server sending one

code/messaging/src/main/java/com/adobe/marketing/mobile/aepcomposeui/components/SmallImageCard.kt
spoorthipujariadobe: Actually bit confused. Does both `UIEvent.Interact` and `UIAction.Click` have an id? Do we need it in both places?

code/messaging/src/main/java/com/adobe/marketing/mobile/aepcomposeui/utils/UIAction.kt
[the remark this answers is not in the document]
rymorale: i'll make the id a required param by using spoorthi's suggestion of using "dismiss" as an id for dismiss button and template id for content cards.

code/messaging/src/main/java/com/adobe/marketing/mobile/aepcomposeui/components/SmallImageCard.kt
rymorale: we do not need it in both places. if we have the id already in UIAction then I think we are good.

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardMapper.kt
```diff
@@ -0,0 +1,48 @@
...
+    private val contentCardSchemaDataMap: MutableMap<String, ContentCardSchemaData> = HashMap()
+
+    companion object {
+        @JvmStatic
+        val instance: ContentCardMapper by lazy { Holder.INSTANCE }
+    }
+
+    private object Holder {
+        val INSTANCE = ContentCardMapper()
+    }
```
prudrabhat, line 29: Is it possible to remove the `Holder` and use `INSTANCE` from companion object directly?
```
companion object {
   val INSTANCE :  ContentCardMapper by lazy { ContentCardMapper() }
}
```

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardMapper.kt
```diff
@@ -0,0 +1,48 @@
...
+    fun removeContentCardSchemaDataFromMap(propositionId: String) {
+        contentCardSchemaDataMap.remove(propositionId)
+    }
+
+    fun getContentCardSchemaDataForPropositionId(propositionId: String): ContentCardSchemaData? {
+        if (StringUtils.isNullOrEmpty(propositionId)) {
+            return null
+        }
+        return contentCardSchemaDataMap[propositionId]
+    }
```
prudrabhat, line 47: Are these public API's we intend to expose from Messaging? If so, can you please add docs about how and when to use them.

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/MessagingEventHandler.kt
```diff
@@ -0,0 +1,44 @@
...
+            callback?.onDisplay(event.aepUi)
+            eventType = MessagingEdgeEventType.DISPLAY
+        } else if (event is UIEvent.Dismiss) {
+            Log.trace(MessagingConstants.LOG_TAG, SELF_TAG, "${event.aepUi.getTemplate().getType()} Dismissed")
+            callback?.onDismiss(event.aepUi)
+            eventType = MessagingEdgeEventType.DISMISS
+        }
+
+        val contentCardSchemaData = ContentCardMapper.instance.getContentCardSchemaDataForPropositionId(propositionId)
+        contentCardSchemaData?.track(null, eventType)
```
prudrabhat, line 42: We should only track from the super class of  Dismiss and Display events. While the code flow takes care of this now, a possible implementation issue in a subclass can result in tracking null eventTypes as well(if a call is made from subclass for non-display/dismiss events).
Can we move these two lines into a private utility and call it individually inside the clauses.

prudrabhat (approved):

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardMapper.kt
rymorale: i'll get docs added for these

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/MessagingEventHandler.kt
rymorale: sure, will get these moved

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/MessagingEventHandlerUtils.kt
```diff
@@ -0,0 +1,27 @@
...
+ */
+internal class MessagingEventHandlerUtils {
+    companion object {
+        @JvmStatic
+        internal fun track(propositionId: String, interaction: String?, eventType: MessagingEdgeEventType?) {
+            val contentCardSchemaData = ContentCardMapper.instance.getContentCardSchemaDataForPropositionId(propositionId)
+            contentCardSchemaData?.track(interaction, eventType)
+        }
+    }
+}
```
prudrabhat, line 27: Why not have this in the companion of MessagingEventHandler as a internal method?

rymorale: My bad, I’ll move it inside

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardMapper.kt
```diff
@@ -0,0 +1,60 @@
...
+        contentCardSchemaDataMap.remove(propositionId)
+    }
+
+    /**
+     * Returns the [ContentCardSchemaData] for the given proposition id.
+     *
+     * @param propositionId the proposition id to retrieve the [ContentCardSchemaData] for
+     * @return the [ContentCardSchemaData] for the given proposition id, or null if not found
+     */
+    fun getContentCardSchemaDataForPropositionId(propositionId: String): ContentCardSchemaData? {
```
prudrabhat, line 54: Was chatting with Spoorthi and it seems like we only need this utility method to be public. Can we rename this to
`fun getContentCardSchemaData(propositionId: String)` and make only this public and switch rest of them to internal.

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardMapper.kt
```diff
@@ -0,0 +1,60 @@
...
+        @JvmStatic
+        val instance: ContentCardMapper by lazy { ContentCardMapper() }
+    }
+
+    /**
+     * Stores the [ContentCardSchemaData] for the given proposition id.
+     *
+     * @param contentCardSchemaData the [ContentCardSchemaData] to store
+     */
+    fun storeContentCardSchemaData(contentCardSchemaData: ContentCardSchemaData) {
```
spoorthipujariadobe, line 32: Can we make this internal so only Messaging classes can change the map stored here?

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardMapper.kt
```diff
@@ -0,0 +1,60 @@
...
+        }
+        contentCardSchemaDataMap[contentCardSchemaData.parent.proposition.uniqueId] = contentCardSchemaData
+    }
+
+    /**
+     * Removes the [ContentCardSchemaData] for the given proposition id.
+     *
+     * @param propositionId the proposition id to remove the [ContentCardSchemaData] for
+     */
+    fun removeContentCardSchemaDataFromMap(propositionId: String) {
```
spoorthipujariadobe, line 44: Same comment here. Let's make this internal

PR #292, opened 2024-10-23
Content provider

navratan-soni:
## Description

Added ContentCardUiProvider class:
- Provides content flow management for UI templates.
- Fetches and processes propositions from surfaces asynchronously.
Added ContentCardSchemaDataUtils object:
- Utility functions for parsing data and creating model instances from a map.
Enhanced KDocs for readability and better documentation of all major classes and methods.

## Types of changes

## Checklist:


code/messaging/src/main/java/com/adobe/marketing/mobile/aepcomposeui/uimodels/AepButton.kt
```diff
@@ -17,16 +17,9 @@ package com.adobe.marketing.mobile.aepcomposeui.uimodels
...
  * @property actionUrl The URL to be opened when the button is clicked.
  * @property text The text to be displayed on the button, represented by an [AepText] object.
- * @property borderWidth The border width of the button.
- * @property borderColor The border color of the button, represented by an [AepColor] object.
- * @property backgroundColour The background color of the button, represented by an [AepColor] object.
- *
  */
 data class AepButton(
-    val id: String,
+    val interactId: String,
```
spoorthipujariadobe, line 22: Please leave this as `id` to keep it generic

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardSchemaData.java
```diff
@@ -69,6 +72,10 @@ public Object getContent() {
         return content;
     }
 
+    public String getContentJsonString() {
```
spoorthipujariadobe, line 75: Can we not expose this as a public API? Why will customers need to use the string content?

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardSchemaData.java
```diff
@@ -39,6 +40,8 @@ public class ContentCardSchemaData implements SchemaData {
                             schemaData.optString(
                                     MessagingConstants.ConsequenceDetailDataKeys.CONTENT_TYPE));
             if (contentType.equals(ContentType.APPLICATION_JSON)) {
+                this.contentJsonString =
+                        schemaData.getString(MessagingConstants.ConsequenceDetailDataKeys.CONTENT);
                 this.content =
                         JSONUtils.toMap(
```
spoorthipujariadobe, line 46: Since we're already converting to map here, wouldn't it be more efficient to use the map rather than convert it to json again and read the keys? Isn't read from map faster than json?

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/contentcards/AepModelUtils.kt
```diff
@@ -0,0 +1,90 @@
...
+import com.adobe.marketing.mobile.aepcomposeui.uimodels.AepColor
+import com.adobe.marketing.mobile.aepcomposeui.uimodels.AepDismissButton
+import com.adobe.marketing.mobile.aepcomposeui.uimodels.AepImage
+import com.adobe.marketing.mobile.aepcomposeui.uimodels.AepText
+import com.adobe.marketing.mobile.aepcomposeui.uimodels.AepUITemplate
+import com.adobe.marketing.mobile.aepcomposeui.uimodels.SmallImageTemplate
+import com.adobe.marketing.mobile.messaging.ContentCardSchemaData
+import org.json.JSONObject
+
+object AepModelUtils {
```
spoorthipujariadobe, line 25: can we rename this to `ContentCardUtils` or `ContentCardSchemaDataUtils`? Also while you have this `contentcards` folder created, wondering if you want to move all existing messaging content cards related classes in here as well

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardSchemaData.java
spoorthipujariadobe: Another advantage of using map is you can directly use `DataReader` provided by Core that already does a lot of null checks and error handling

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/contentcards/ContentCardUiProvider.kt
```diff
@@ -0,0 +1,148 @@
...
+                        Log.error(MessagingConstants.LOG_TAG, SELF_TAG, "Failed to get content: ${error.message}")
+                        _contentFlow.value = emptyList()
+                    }
+                }
+            }
+        }
+        return contentFlow
+    }
+
+    fun getCardsForSurface(
```
spoorthipujariadobe, line 60: can we make this `private`?

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/contentcards/ContentCardUiProvider.kt
```diff
@@ -0,0 +1,148 @@
...
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.launch
+
+class ContentCardUiProvider(val surfaceString: String) : AepUIContentProvider {
+    companion object {
+        private const val SELF_TAG: String = "MessageAssetDownloader"
```
spoorthipujariadobe, line 33: SELF_TAG needs to be `ContentCardUiProvider`

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/contentcards/ContentCardUiProvider.kt
```diff
@@ -0,0 +1,148 @@
...
+
+    private val _contentFlow = MutableStateFlow<List<AepUITemplate>>(emptyList())
+    private val contentFlow: StateFlow<List<AepUITemplate>> = _contentFlow
+
+    override suspend fun getContent(): Flow<List<AepUITemplate>> {
+        CoroutineScope(Dispatchers.IO).launch {
+            val surface = Surface(surfaceString)
+            val surfaceList = mutableListOf<Surface>()
+            surfaceList.add(surface)
+            Messaging.updatePropositionsForSurfaces(surfaceList)
```
spoorthipujariadobe, line 44: Why are we calling `updatePropositionsForSurfaces` and `getPropositionsForSurfaces` here as well as inside `getCardsForSurface`? Can we extract this code into a common method and use in both places?

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/contentcards/ContentCardUiProvider.kt
```diff
@@ -0,0 +1,148 @@
...
+import com.adobe.marketing.mobile.messaging.Surface
+import com.adobe.marketing.mobile.services.Log
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.launch
+
+class ContentCardUiProvider(val surfaceString: String) : AepUIContentProvider {
```
spoorthipujariadobe, line 31: Please add kdocs for this class

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/contentcards/ContentCardUiProvider.kt
```diff
@@ -0,0 +1,148 @@
...
+    companion object {
+        private const val SELF_TAG: String = "MessageAssetDownloader"
+    }
+
+    private val _contentFlow = MutableStateFlow<List<AepUITemplate>>(emptyList())
+    private val contentFlow: StateFlow<List<AepUITemplate>> = _contentFlow
+
+    override suspend fun getContent(): Flow<List<AepUITemplate>> {
```
spoorthipujariadobe, line 39: Please add kdocs for this class

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/contentcards/ContentCardUiProvider.kt
spoorthipujariadobe, line 60: Please add kdocs for this

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/contentcards/ContentCardUiProvider.kt
```diff
@@ -0,0 +1,148 @@
...
+                        _contentFlow.value = emptyList()
+                    }
+                }
+            }
+        }
+        return contentFlow
+    }
+
+    fun getCardsForSurface(
+        surfaceList: MutableList<Surface>,
```
spoorthipujariadobe, line 61: Why do we need a separate `surfaceList` here? The class construction already accepts a surface

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/contentcards/ContentCardUiProvider.kt
```diff
@@ -0,0 +1,148 @@
...
+            }
+        )
+    }
+
+    override suspend fun refreshContent() {
+        _contentFlow.value = emptyList()
+        getContent()
+    }
+
+    fun buildTemplate(proposition: Proposition): AepUITemplate? {
```
spoorthipujariadobe, line 134: can we make this `private`?

spoorthipujariadobe (changes requested):
1. Please add a description to the PR on what it includes
2. Please fix the CI failures

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/contentcards/ContentCardUiProvider.kt
spoorthipujariadobe: Can you rename this to `getAepUi`? The description of what needs to be done is in this ticket https://jira.corp.adobe.com/browse/MOB-22204

code/messaging/src/main/java/com/adobe/marketing/mobile/aepcomposeui/uimodels/AepImage.kt
```diff
@@ -16,23 +16,8 @@ package com.adobe.marketing.mobile.aepcomposeui.uimodels
...
  */
 data class AepImage(
     val url: String? = null,
     val darkUrl: String? = null,
-    val bundle: String? = null,
-    val darkBundle: String? = null,
-    // TODO: verify Android equivalent of icon
-    val icon: String? = null,
-    val iconSize: Float? = null,
-    val iconColor: AepColor? = null,
```
spoorthipujariadobe, line 35: Can you also delete AepColor if its not used? We can add it when server sends it

code/messaging/src/main/java/com/adobe/marketing/mobile/aepcomposeui/AepUIConstants.kt
```diff
@@ -26,4 +26,22 @@ internal object AepUIConstants {
             const val SPACING = 8
         }
     }
+
+    internal object Keys {
```
spoorthipujariadobe, line 30: Can we move these to MessagingConstants? These are not used by UI library right?

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/contentcards/ContentCardUiProvider.kt
navratan-soni: Messaging update and get proposition needs List<Surface> as input, so need to wrap one surface as list as well.

spoorthipujariadobe: Surface is a location in the app where cards are to be displayed. While it makes sense to pass a list to prefetch and cache multiple surfaces in updatePropositionsForSurface since it makes the actual network call, why would customers need to get more than a single surface while actually displaying? Wouldn't they display one location at a time? How will they handle a list?

code/messaging/src/main/java/com/adobe/marketing/mobile/aepcomposeui/contentprovider/AepUIContentProvider.kt
```diff
@@ -21,9 +21,9 @@ import kotlinx.coroutines.flow.Flow
 interface AepUIContentProvider {
     /**
      * Retrieves the content for the UI.
-     * @return The content for the UI as a flow of [AepTemplate]s.
+     * @return The content for the UI as a flow of [AepUI]s.
      */
-    suspend fun getContent(): Flow<List<AepUITemplate>>
+    suspend fun getContent(): Flow<List<AepUI<*, *>>>
```
spoorthipujariadobe, line 26: Why did we change this interface? It needs to be as is to be used by container, you can add an additional function in Messaging implementation if you want

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/MessagingConstants.java
```diff
@@ -407,4 +407,20 @@ public class PayloadKeys {
             private PayloadKeys() {}
         }
     }
+
+    public final class AepUIKeys {
```
spoorthipujariadobe, line 411: Can you name these UIKeys to be consistent with the rest of the classes in this file? Also its specific to messaging, not sure why we need `Aep` prefix here

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/contentcards/ContentCardSchemaDataUtils.kt
```diff
@@ -0,0 +1,138 @@
...
+            id = id,
+            title = title,
+            body = body,
+            image = image,
+            actionUrl = actionUrl,
+            buttons = buttons,
+            dismissBtn = dismissBtn
+        )
+    } catch (e: Exception) {
+        throw IllegalArgumentException("Error parsing content map: ${e.message}")
```
spoorthipujariadobe, line 120: If we use DataReader.opt* we can avoid throwing this exception

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/contentcards/ContentCardSchemaDataUtils.kt
```diff
@@ -0,0 +1,138 @@
...
+ * @param contentCardSchemaData The content card schema data containing the content map.
+ * @return The parsed [AepUITemplate] or null if parsing fails.
+ * @throws IllegalArgumentException If the content map is null or if there is an error during parsing.
+ */
+fun getTemplateModelFromContentCardSchemaData(contentCardSchemaData: ContentCardSchemaData): AepUITemplate? {
+    val contentMap =
+        contentCardSchemaData.content as? Map<String, Any> ?: throw IllegalArgumentException("Content map is null")
+    try {
+        val id = contentMap[MessagingConstants.AepUIKeys.ID] as String
+        val title =
```
spoorthipujariadobe, line 95: `title` is required as per spec. If it is null, empty or not provided can you return a null `AepUITemplate` here

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/contentcards/ContentCardSchemaDataUtils.kt
```diff
@@ -0,0 +1,138 @@
...
+}
+
+/**
+ * Parses the given [ContentCardSchemaData] to create an [AepUITemplate].
+ *
+ * @param contentCardSchemaData The content card schema data containing the content map.
+ * @return The parsed [AepUITemplate] or null if parsing fails.
+ * @throws IllegalArgumentException If the content map is null or if there is an error during parsing.
+ */
+fun getTemplateModelFromContentCardSchemaData(contentCardSchemaData: ContentCardSchemaData): AepUITemplate? {
```
spoorthipujariadobe, line 90: nit: can we remove the part after `From` and call it `getTemplateModel`? The from is implicit based on the argument you're passing in

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/contentcards/ContentCardSchemaDataUtils.kt
```diff
@@ -0,0 +1,138 @@
...
+ * @param uiTemplate The template to convert into a UI component.
+ * @return An instance of [AepUI] representing the given template.
+ * @throws IllegalArgumentException If the provided template type is unsupported.
+ */
+fun provideAepUIfromAepUITemplate(uiTemplate: AepUITemplate): AepUI<*, *> {
+    return when (uiTemplate) {
+        is SmallImageTemplate -> {
+            SmallImageUI(uiTemplate, SmallImageCardUIState())
+        }
+        else -> throw IllegalArgumentException("Unsupported template type")
```
spoorthipujariadobe, line 136: Same comment here. Where is the exception caught? Can we handle this gracefully?

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/contentcards/ContentCardUiProvider.kt
```diff
@@ -0,0 +1,181 @@
...
+     * @param surfaceList The list of surfaces for which to retrieve the templates.
+     * @param completion The callback to handle the result of the template retrieval.
+     */
+
+    private suspend fun getAepUIForSurfaceList(
+        surfaceList: MutableList<Surface>,
+        completion: (Result<List<AepUITemplate>>) -> Unit
+    ) {
+        // Update propositions for the provided surface
+        Messaging.updatePropositionsForSurfaces(surfaceList)
```
spoorthipujariadobe, line 87: Can we remove the `updatePropositionsForSurfaces` here? Customers ideally need to call this well in advance before the UI is being displayed. Also aligns with iOS implementation

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/contentcards/ContentCardUiProvider.kt
```diff
@@ -0,0 +1,181 @@
...
+     * Retrieves the list of [AepUITemplate] for the provided surfaces.
+     * Updates the propositions and retrieves templates based on those propositions.
+     *
+     * @param surfaceList The list of surfaces for which to retrieve the templates.
+     * @param completion The callback to handle the result of the template retrieval.
+     */
+
+    private suspend fun getAepUIForSurfaceList(
+        surfaceList: MutableList<Surface>,
+        completion: (Result<List<AepUITemplate>>) -> Unit
```
spoorthipujariadobe, line 84: Can we keep method signatures consistent here? One returns a `Flow` another uses a callback. Can we please stick to using Flow and make this a suspend function?

Also why do we need both `getAepUIForSurfaceList` and `provideAepUIfromAepUITemplate`? Aren't they doing the same thing? Can we keep one of them?

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/contentcards/ContentCardUiProvider.kt
navratan-soni: get Cards for surface is not public API, renamed to getAepUIForSurfaceList, we are using this function internally.
ContentCardUiProvider takes a single surface as parameter. Then getContent() call is placed by user, we are
we are wrapping this surface as a single element in list. as Messaeging's get and update proposition requires List

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/contentcards/ContentCardUiProvider.kt
navratan-soni: Done

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/contentcards/ContentCardSchemaDataUtils.kt
navratan-soni: Removed this function from scope of this PR

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/contentcards/ContentCardSchemaDataUtils.kt
navratan-soni: Done

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/contentcards/ContentCardSchemaDataUtils.kt
```diff
@@ -0,0 +1,138 @@
...
+}
+
+/**
+ * Provides an [AepUI] instance from the given [AepUITemplate].
+ *
+ * @param uiTemplate The template to convert into a UI component.
+ * @return An instance of [AepUI] representing the given template.
+ * @throws IllegalArgumentException If the provided template type is unsupported.
+ */
+fun provideAepUIfromAepUITemplate(uiTemplate: AepUITemplate): AepUI<*, *> {
```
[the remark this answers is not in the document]
navratan-soni: This is removed from scope of this PR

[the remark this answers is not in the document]
navratan-soni: getAepUi() will add in separate class, these content provider class, will be used by that function in Messaging.

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/contentcards/ContentCardSchemaDataUtils.kt
navratan-soni: Removed from the scope of this PR

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/contentcards/ContentCardUiProvider.kt
navratan-soni: Removed

navratan-soni: This is not a used facing method, but an internal method to be used by getContent() method which will be called by getAepUI()

code/messaging/src/main/java/com/adobe/marketing/mobile/aepcomposeui/components/AepTextComposable.kt
```diff
@@ -108,12 +103,12 @@ internal fun AepTextComposablePreview() {
...
-        AepTextComposable(AepText(content = "Null Font", font = null))
+        AepTextComposable(AepText(content = "Null Color"))
+        AepTextComposable(AepText(content = "Null Alignment"))
+        AepTextComposable(AepText(content = "Null Font"))
 
         // Extreme font sizes
-        AepTextComposable(AepText(content = "Tiny Text", font = AepFont(size = 1)))
-        AepTextComposable(AepText(content = "Huge Text", font = AepFont(size = 100)))
+        AepTextComposable(AepText(content = "Tiny Text"))
+        AepTextComposable(AepText(content = "Huge Text"))
```
spoorthipujariadobe, line 112: Out of the scope of this PR: You can use `AepTextStyle` to create these previews in a following PR

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/MessagingConstants.java
```diff
@@ -407,4 +407,26 @@ public class PayloadKeys {
             private PayloadKeys() {}
         }
     }
+
+    public final class UIKeys {
```
spoorthipujariadobe, line 411: Messaging extension has other UIs as well: Push, IAM. These are specific to content cards. Can you add this inside `public final class ContentCard`

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/MessagingConstants.java
```diff
@@ -407,4 +407,26 @@ public class PayloadKeys {
...
+        public static final String TEXT = "text";
+        public static final String STYLE = "style";
+        public static final String ID = "id";
+        public static final String TITLE = "title";
+        public static final String BODY = "body";
+        public static final String IMAGE = "image";
+        public static final String BUTTONS = "buttons";
+        public static final String DISMISS_BTN = "dismissBtn";
+
+        public class DismissButtonStyle {
```
spoorthipujariadobe, line 426: Can we keep the class hierarchy consistent? If we plan to have everything inside `UIKeys` lets remove this. If not, lets make sure all keys are under respective classes like `Text`, `Button` and `Image`

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/contentcards/ContentCardUiProvider.kt
```diff
@@ -0,0 +1,156 @@
...
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.launch
+
+/**
+ * ContentCardUiProvider is responsible for fetching and managing the content for a given surface.
+ * It uses Adobe Messaging APIs to retrieve propositions and transform them into UI templates for display.
+ *
+ * @property surfaceString The identifier of the surface for which content needs to be fetched.
+ */
+class ContentCardUiProvider(val surfaceString: String) : AepUIContentProvider {
```
spoorthipujariadobe, line 37: Ui needs to be capitalized -> `ContentCardUIProvider`

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/contentcards/ContentCardUiProvider.kt
```diff
@@ -0,0 +1,156 @@
...
+    companion object {
+        private const val SELF_TAG: String = "ContentCardUiProvider"
```
spoorthipujariadobe, line 39: Same comment here `ContentCardUIProvider`

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/contentcards/ContentCardUiProvider.kt
navratan-soni: changed to getAepUITemplateList which is not public API, it is internal function.
getAepUi() and getContent() are two available public APIs

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/contentcards/ContentCardSchemaDataUtils.kt
navratan-soni: Returned null in case of parsing fails.

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/contentcards/ContentCardSchemaDataUtils.kt
navratan-soni: Return null when unsupported type

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/contentcards/ContentCardSchemaDataUtils.kt
```diff
@@ -0,0 +1,138 @@
...
+ * Provides functions to create [AepText], [AepImage], [AepButton], and [AepDismissButton].
+ */
+object ContentCardSchemaDataUtils {
+    /**
+     * Creates an instance of [AepText] from a map of key-value pairs.
+     *
+     * @param map The map containing the data for [AepText].
+     * @return An instance of [AepText].
+     */
+    fun createAepText(map: Map<String, Any>): AepText {
```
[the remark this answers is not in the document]
navratan-soni: To create AepText.from(map:...) and call, we need extension functions here as AepText.COmpanion.from(map:...), and declare empty companion objects in individual data classes.
For now, I have made existing function as private.

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/contentcards/ContentCardUiProvider.kt
```diff
@@ -0,0 +1,181 @@
...
+     * Builds an [AepUITemplate] from a given [Proposition].
+     *
+     * @param proposition The proposition to build the template from.
+     * @return The built [AepUITemplate] or null if the proposition is not a content card.
+     */
+    private fun buildTemplate(proposition: Proposition): AepUITemplate? {
+        var baseTemplateModel: AepUITemplate? = null
+        if (isContentCard(proposition)) {
+            val propositionItem = proposition.items[0]
+            baseTemplateModel = propositionItem.contentCardSchemaData?.let {
```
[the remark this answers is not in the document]
navratan-soni: Method is expecting a non-null proposition value, however at top level there is fail safe for null proposition, creates, empty List

code/messaging/src/main/java/com/adobe/marketing/mobile/aepcomposeui/components/AepTextComposable.kt
navratan-soni: Sure

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/contentcards/ContentCardUiProvider.kt
spoorthipujariadobe: Hmm I still don't get why we need to accept `surfaceList` here. `surfaceString` is passed in when the class is created. Why can't we `surfaceString` directly instead of accepting another list here?

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/contentcards/ContentCardSchemaDataUtils.kt
```diff
@@ -0,0 +1,177 @@
...
+                id = id,
+                title = title,
+                body = body,
+                image = image,
+                actionUrl = actionUrl,
+                buttons = buttons,
+                dismissBtn = dismissBtn
+            )
+        } catch (e: Exception) {
+            return null
```
rymorale, line 128: we should log the exception here for future investigations / debugging.

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/contentcards/ContentCardSchemaDataUtils.kt
```diff
@@ -0,0 +1,177 @@
...
+     * Parses the given [ContentCardSchemaData] to create an [AepUITemplate].
+     *
+     * @param contentCardSchemaData The content card schema data containing the content map.
+     * @return The parsed [AepUITemplate] or null if parsing fails.
+     */
+    internal fun getTemplateModel(contentCardSchemaData: ContentCardSchemaData): AepUITemplate? {
+        try {
+            val contentMap =
+                contentCardSchemaData.content as? Map<String, Any> ?: throw IllegalArgumentException("Content map is null")
+            val id = DataReader.getString(contentMap, MessagingConstants.ContentCard.UIKeys.ID)
```
spoorthipujariadobe, line 97: There is not id field send from the server. This needs to be the proposition id.
`contentCardSchemaData.parent.proposition.uniqueId`

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/contentcards/ContentCardSchemaDataUtils.kt
```diff
@@ -0,0 +1,177 @@
...
+
+            val title = DataReader.getTypedMap(Any::class.java, contentMap, MessagingConstants.ContentCard.UIKeys.TITLE)?.let {
+                createAepText(it)
+            }
+            if (title == null || title.content.isEmpty()) {
+                return null
+            }
+            val body = DataReader.optTypedMap(Any::class.java, contentMap, MessagingConstants.ContentCard.UIKeys.BODY, emptyMap())?.let {
```
spoorthipujariadobe, line 105: The last value to `optTypedMap` passed here is `emptyMap()`, so the let part will always get executed, even if body does not exist in the server response. Did you want to pass null instead?

Same question with image, actionUrl and buttons

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/contentcards/ContentCardSchemaDataUtils.kt
```diff
@@ -0,0 +1,177 @@
...
+internal object ContentCardSchemaDataUtils {
+    /**
+     * Creates an instance of [AepText] from a map of key-value pairs.
+     *
+     * @param map The map containing the data for [AepText].
+     * @return An instance of [AepText].
+     */
+    private fun createAepText(map: Map<String, Any>): AepText {
+        return AepText(
+            content = DataReader.optString(map, MessagingConstants.ContentCard.UIKeys.CONTENT, "")
```
spoorthipujariadobe, line 44: We're not really enforcing `content` being mandatory for `AepText` here. Would you please add that check? Maybe something like this?
```
DataReader.optString(map, MessagingConstants.ContentCard.UIKeys.CONTENT, null)?.let {
     AepText(content = it)
}
```

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/contentcards/ContentCardSchemaDataUtils.kt
```diff
@@ -0,0 +1,177 @@
...
+     * Creates an instance of [AepButton] from a map of key-value pairs.
+     *
+     * @param map The map containing the data for [AepButton].
+     * @return An instance of [AepButton].
+     */
+    private fun createAepButton(map: Map<String, Any>): AepButton {
+        return AepButton(
+            id = DataReader.optString(map, MessagingConstants.ContentCard.UIKeys.INTERACT_ID, null),
+            actionUrl = DataReader.optString(map, MessagingConstants.ContentCard.UIKeys.ACTION_URL, null),
+            text = DataReader.optTypedMap(Any::class.java, map, MessagingConstants.ContentCard.UIKeys.TEXT, null).let { createAepText(it) }
```
spoorthipujariadobe, line 71: id, actionUrl and text for a `AepButton` cannot be null. This will lead to runtime exception. You might need to check they are non null before creating `AepButton`

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/contentcards/ContentCardSchemaDataUtils.kt
```diff
@@ -0,0 +1,177 @@
...
+                createAepImage(it)
+            }
+            val actionUrl = DataReader.optString(contentMap, MessagingConstants.ContentCard.UIKeys.ACTION_URL, null)
+            val buttons = DataReader.optTypedListOfMap(Any::class.java, contentMap, MessagingConstants.ContentCard.UIKeys.BUTTONS, emptyList()).map {
+                createAepButton(it)
+            }
+            val dismissBtn = DataReader.optTypedMap(Any::class.java, contentMap, MessagingConstants.ContentCard.UIKeys.DISMISS_BTN, emptyMap())?.let {
+                createAepDismissButton(it)
+            }
+            return SmallImageTemplate(
```
spoorthipujariadobe, line 118: The `template` type is inside the meta data field. You should be using that to decide which type of template to create. Please refer to this spec for more details
https://wiki.corp.adobe.com/display/adms/Small+Image+Template

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/contentcards/ContentCardUIProvider.kt
```diff
@@ -0,0 +1,163 @@
...
+        // Retrieve propositions for the provided surface
+        Messaging.getPropositionsForSurfaces(
+            surfaceList,
+            object :
+                AdobeCallbackWithError<Map<Surface, List<Proposition>>> {
+                override fun call(resultMap: Map<Surface, List<Proposition>>?) {
+                    if (resultMap == null) {
+                        completion(
+                            Result.failure(
+                                Throwable(
```
spoorthipujariadobe, line 115: Please add a log statement here with reason for failure

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/contentcards/ContentCardUIProvider.kt
```diff
@@ -0,0 +1,163 @@
...
+                    _contentFlow.value = emptyList()
+                    _aepUiFlow.value = emptyList()
+                }
+            }
+        }
+        return contentFlow
+    }
+
+    private suspend fun getAepUITemplateList(
+        surfaceList: MutableList<Surface>,
```
spoorthipujariadobe, line 103: Still not sure why this needs to be a list. Please refer to this comment
https://github.com/adobe/aepsdk-messaging-android/pull/292#discussion_r1819689273

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/contentcards/ContentCardUiProvider.kt
navratan-soni: Done

code/messaging/src/main/java/com/adobe/marketing/mobile/aepcomposeui/uimodels/AepImage.kt
```diff
@@ -16,23 +16,8 @@ package com.adobe.marketing.mobile.aepcomposeui.uimodels
...
- * @property darkBundle The resource bundle for the image in dark mode.
- * @property icon The icon name or identifier.
- * @property iconSize The size of the icon.
- * @property iconColor The color of the icon.
- * @property alt The alt text for the image.
- * @property placeholder The placeholder image for the image.
  */
 data class AepImage(
-    val url: String? = null,
+    val url: String,
```
spoorthipujariadobe, line 21: Why did we change this to non null? As per spec its optional
https://wiki.corp.adobe.com/display/adms/AEPImage

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardSchemaDataUtils.kt
```diff
@@ -0,0 +1,213 @@
...
+    const val SELF_TAG = "ContentCardSchemaDataUtils"
+    /**
+     * Creates an instance of [AepText] from a map of key-value pairs.
+     *
+     * @param map The map containing the data for [AepText].
+     * @return An instance of [AepText].
+     */
+    private fun createAepText(map: Map<String, Any>): AepText {
+        return AepText(
+            content = DataReader.optString(map, MessagingConstants.ContentCard.UIKeys.CONTENT, "")
```
spoorthipujariadobe, line 43: nit: Can we also assign this to `null` for
1. Consistency with other methods
2. Causes exception if an AepText is created without `content` which should is erroneous and ensures template is null

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardSchemaDataUtils.kt
```diff
@@ -0,0 +1,213 @@
...
+                createAepDismissButton(it)
+            }
+
+            val metaAdobeObject =
+                contentCardSchemaData.meta?.get(MessagingConstants.ContentCard.UIKeys.ADOBE)
+            metaAdobeObject.let {
+                val templateType =
+                    (it as? Map<String, Any>)?.get(MessagingConstants.ContentCard.UIKeys.TEMPLATE)
+                when (templateType) {
+                    AepUITemplateType.SMALL_IMAGE.typeName -> {
```
spoorthipujariadobe, line 132: Can we move this check to L100, before reading the card fields? title, body, image etc. are specific to SmallImageTemplate and should be read based on conforming template type

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardUIProvider.kt
```diff
@@ -0,0 +1,160 @@
...
+     *
+     * @return A flow that emits lists of AepUITemplate.
+     */
+    override suspend fun getContent(): Flow<List<AepUITemplate>> {
+        CoroutineScope(Dispatchers.IO).launch {
+            val surface = Surface(surfaceString)
+            getAepUITemplateList(surface) { it ->
+                it.onSuccess { templateList ->
+                    _contentFlow.value = templateList
+                    _aepUiFlow.value = templateList.mapNotNull { item -> getAepUI(item) }
```
spoorthipujariadobe, line 82: nit: Can we move this line to `getAepUi` method? If customer is going to be using `AepList` and isn't going to be calling `getAepUi` does it make sense to maintain a flow of `aepUi`?

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardUIProvider.kt
```diff
@@ -0,0 +1,160 @@
...
+        getContent()
+        return aepUiFlow
+    }
+
+    /**
+     * Clears the current content and re-fetches new content for the given surface.
+     */
+    override suspend fun refreshContent() {
+        _contentFlow.value = emptyList()
+        _aepUiFlow.value = emptyList()
```
spoorthipujariadobe, line 66: The content cards list, and in turn the views, will be empty in the time between requesting for new content and the lists being updated, which isn't great. Can we avoid updating these to empty lists while refreshing content?

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardUIProvider.kt
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    /**
+     * Retrieves a flow of AepUITemplate lists for the given surface.
+     * The flow emits updates whenever new content is fetched.
+     *
+     * @return A flow that emits lists of AepUITemplate.
+     */
+    override suspend fun getContent(): Flow<List<AepUITemplate>> {
+        CoroutineScope(Dispatchers.IO).launch {
```
spoorthipujariadobe, line 77: Do you think we can reuse a single scope stored at class level? Would that be more efficient?
@prudrabhat what do you think?

prudrabhat: Given that this is a suspending function, and the caller will already have to invoke and launch this in a CoroutineScope, we can write the business logic directly. The caller can decide what scope to run this on.

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardUIProvider.kt
```diff
@@ -0,0 +1,160 @@
...
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.launch
+
+/**
+ * ContentCardUiProvider is responsible for fetching and managing the content for a given surface.
+ * It uses Adobe Messaging APIs to retrieve propositions and transform them into UI templates for display.
+ *
+ * @property surfaceString The identifier of the surface for which content needs to be fetched.
+ */
+class ContentCardUIProvider(val surfaceString: String) : AepUIContentProvider {
```
spoorthipujariadobe, line 36: Can we accept a Surface here instead of string? Customer would already have a surface that was used for `updatePropositionsForSurface`

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardUIProvider.kt
```diff
@@ -0,0 +1,160 @@
...
+                    _contentFlow.value = emptyList()
+                    _aepUiFlow.value = emptyList()
+                }
+            }
+        }
+        return contentFlow
+    }
+
+    private suspend fun getAepUITemplateList(
+        surface: Surface,
```
spoorthipujariadobe, line 98: If we accept a surface instead of string in the class constructor we should be able to remove this

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardUIProvider.kt
prudrabhat: I think this is being done to have `contentFlow` and `aepUiFlow` to be in sync with each other incase some one calls  `getContent()`. That makes me think if it is necessary to keep them in syc... may be we need to for maintainability. 

If we need to have them in sync-  does this work?
```
    val aepUiFlow: Flow<List<..>> = _contentFlow.map { templateList ->
        templateList.mapNotNull { item -> getAepUI(item) }
    }
```

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardUIProvider.kt
```diff
@@ -0,0 +1,160 @@
...
+                    _contentFlow.value = templateList
+                    _aepUiFlow.value = templateList.mapNotNull { item -> getAepUI(item) }
+                }
+                it.onFailure { error ->
+                    Log.error(
+                        MessagingConstants.LOG_TAG, SELF_TAG,
+                        "Failed to get content: ${error.message}"
+                    )
+                    _contentFlow.value = emptyList()
+                    _aepUiFlow.value = emptyList()
```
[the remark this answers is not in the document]
spoorthipujariadobe: I was unsure about this one too. It might be better to keep the existing cards if failure occurs

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardSchemaDataUtils.kt
```diff
@@ -0,0 +1,213 @@
...
+        }
+    }
+
+    /**
+     * Parses the given [ContentCardSchemaData] to create an [AepUITemplate].
+     *
+     * @param contentCardSchemaData The content card schema data containing the content map.
+     * @return The parsed [AepUITemplate] or null if parsing fails.
+     */
+    internal fun getTemplateModel(contentCardSchemaData: ContentCardSchemaData): AepUITemplate? {
```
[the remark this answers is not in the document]
navratan-soni: Broken into two parts, getTemplate() changed the method name
1. Identify the Template Type
2. Based on template type detected, parsed the contentMap.

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardUIProvider.kt
```diff
@@ -0,0 +1,160 @@
...
+    /**
+     * Retrieves a flow of AepUI instances for the given surface.
+     *
+     * This function initiates the content fetch using [getContent] and then returns a flow of
+     * [AepUI] instances that represent the UI templates. The flow emits updates whenever new
+     * content is fetched or any changes occur.
+     *
+     * @return A [Flow] that emits a list of [AepUI] instances.
+     */
+    suspend fun getAepUi(): Flow<List<AepUI<*, *>>> {
```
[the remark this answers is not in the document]
navratan-soni: I believe this is the name we kept in sync with iOS only.

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardSchemaDataUtils.kt
```diff
@@ -0,0 +1,213 @@
...
+
+    /**
+     * Builds an [AepUITemplate] from a given [Proposition].
+     *
+     * @param proposition The proposition containing the content card data.
+     * @return The built [AepUITemplate] or null if the proposition is not a content card or parsing fails.
+     */
+    internal fun buildTemplate(proposition: Proposition): AepUITemplate? {
+        var baseTemplateModel: AepUITemplate? = null
+        if (isContentCard(proposition)) {
```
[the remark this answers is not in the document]
navratan-soni: Cleaned up the code.

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardUIProvider.kt
navratan-soni: Done

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardUIProvider.kt
navratan-soni: Done

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardUIProvider.kt
navratan-soni: Removed, done

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardUIProvider.kt
navratan-soni: Changed the flow. Done

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardSchemaDataUtils.kt
navratan-soni: Moved the check, and parsing based on template is moved to a separate function to keep this function body small.

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardSchemaDataUtils.kt
navratan-soni: Done

code/messaging/src/main/java/com/adobe/marketing/mobile/aepcomposeui/uimodels/AepImage.kt
navratan-soni: Done

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardUIProvider.kt
```diff
@@ -0,0 +1,149 @@
...
+                    }
+
+                    val templateModelList = mutableListOf<AepUITemplate>()
+                    Log.debug(
+                        MessagingConstants.LOG_TAG,
+                        SELF_TAG,
+                        "getPropositionsForSurfaces callback contained Null Map"
+                    )
+
+                    for ((_, propositions) in resultMap.entries) {
```
spoorthipujariadobe, line 116: nit: Do we need the nested for loop here? Does `val propositions: resultMap[surface]` followed by for-in not work?

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardSchemaDataUtils.kt
```diff
@@ -0,0 +1,220 @@
...
+        try {
+            val contentMap =
+                contentCardSchemaData.content as? Map<String, Any> ?: throw IllegalArgumentException("Content map is null")
+            val id = contentCardSchemaData.parent.proposition.uniqueId
+
+            val metaAdobeObject =
+                contentCardSchemaData.meta?.get(MessagingConstants.ContentCard.UIKeys.ADOBE)
+
+            metaAdobeObject.let { mapObject ->
+                val templateType =
```
spoorthipujariadobe, line 105: nit: We can fast fail here if `metaAdobeObject` is null or empty

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardSchemaDataUtils.kt
```diff
@@ -0,0 +1,220 @@
...
+    internal fun getTemplate(contentCardSchemaData: ContentCardSchemaData): AepUITemplate? {
+        try {
+            val contentMap =
+                contentCardSchemaData.content as? Map<String, Any> ?: throw IllegalArgumentException("Content map is null")
+            val id = contentCardSchemaData.parent.proposition.uniqueId
+
+            val metaAdobeObject =
+                contentCardSchemaData.meta?.get(MessagingConstants.ContentCard.UIKeys.ADOBE)
+
+            metaAdobeObject.let { mapObject ->
```
spoorthipujariadobe, line 104: nit: We can fast fail here if `metaAdobeObject` is null or empty

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardSchemaDataUtils.kt
```diff
@@ -0,0 +1,220 @@
...
+                val templateType =
+                    (mapObject as? Map<String, Any>)?.get(MessagingConstants.ContentCard.UIKeys.TEMPLATE)
```
spoorthipujariadobe, line 106: nit: Can we avoid casting by using DataReader?

spoorthipujariadobe (approved):
Couple of nits, looks fine overall

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardUIProvider.kt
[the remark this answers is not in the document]
navratan-soni: Changed to getContentCardUI() as per iOS

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardUIProvider.kt
navratan-soni: Done

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardSchemaDataUtils.kt
navratan-soni: Will do in next PR for UTs

PR #294, opened 2024-10-30
Changing UI state on SmallImageCard events + Test app updates

spoorthipujariadobe:
## Description

1. Added an `isLoading` flag to be used when image is being downloaded from the url
2. AepUiState `dismissed` is set to `true` when `UIEvent.Dismiss` is received.
3. SmallImageCard composable is recomposed on configuration changes like theme or orientation change leading to multiple `UIEvent.Display` events for a single "view". To ensure `decisioning.propositionDisplay` events are sent only on the initial composition, `displayed` is set to `true` in AepUiState after the first display. `ContentCardSchemaData.track(MessagingEdgeEventType.DISPLAY)` is only called if `displayed` is `false`
4. Updated messaging test app to use content card UI library

## Types of changes

## Checklist:

- [x] I have signed the [Adobe Open Source CLA](http://opensource.adobe.com/cla.html).
- [x] My code follows the code style of this project.
- [x] I have read the **CONTRIBUTING** document.
- [x] All new and existing tests passed.

### DIFF
--- a/code/messaging/src/main/java/com/adobe/marketing/mobile/aepcomposeui/components/SmallImageCard.kt
+++ b/code/messaging/src/main/java/com/adobe/marketing/mobile/aepcomposeui/components/SmallImageCard.kt
@@ -132,7 +132,9 @@ fun SmallImageCard(
                                 onClick = {
                                     observer?.onEvent(UIEvent.Interact(ui, UIAction.Click(button.id, button.actionUrl)))
                                 },
-                                buttonStyle = style.buttonStyle[index].first,
+                                buttonStyle = style.buttonStyle[index].first.apply {
+                                    modifier = (modifier ?: Modifier).then(Modifier.weight(1f))
+                                },
                                 buttonTextStyle = style.buttonStyle[index].second
                             )
                         }
--- a/code/messaging/src/main/java/com/adobe/marketing/mobile/aepcomposeui/components/SmallImageCard.kt
+++ b/code/messaging/src/main/java/com/adobe/marketing/mobile/aepcomposeui/components/SmallImageCard.kt
@@ -58,8 +58,16 @@ fun SmallImageCard(
         if (imageUrl.isNullOrBlank()) {
             isLoading = false
         } else {
-            imageBitmap = UIUtils.downloadImage(imageUrl)
-            isLoading = false
+            UIUtils.downloadImage(imageUrl) {
+                it.onSuccess { bitmap ->
+                    imageBitmap = bitmap
+                    isLoading = false
+                }
+                it.onFailure {
+                    // TODO once we have a default image, we can use that here
+                    isLoading = false
+                }
+            }
         }
     }
 
--- a/code/messaging/src/main/java/com/adobe/marketing/mobile/aepcomposeui/components/SmallImageCard.kt
+++ b/code/messaging/src/main/java/com/adobe/marketing/mobile/aepcomposeui/components/SmallImageCard.kt
@@ -50,16 +50,20 @@ fun SmallImageCard(
     style: SmallImageUIStyle,
     observer: AepUIEventObserver?,
 ) {
+    var isLoading by remember { mutableStateOf(true) }
     var imageBitmap by remember { mutableStateOf<Bitmap?>(null) }
     val imageUrl = if (isSystemInDarkTheme() && ui.getTemplate().image?.darkUrl != null)
         ui.getTemplate().image?.darkUrl else ui.getTemplate().image?.url
 
     LaunchedEffect(ui.getTemplate().id) {
         observer?.onEvent(UIEvent.Display(ui))
-        if (imageUrl != null) {
+        if (imageUrl.isNullOrBlank()) {
+            isLoading = false
+        } else {
             imageBitmap = withContext(Dispatchers.IO) {
                 UIUtils.downloadImage(imageUrl)
             }
+            isLoading = false
         }
     }
 
@@ -91,7 +95,8 @@ fun SmallImageCard(
                         content = BitmapPainter(it.asImageBitmap()),
                         style = style.imageStyle
                     )
-                } ?: run {
+                }
+                if (isLoading) {
                     Box(
                         modifier = Modifier
                             .size(AepUIConstants.SmallImageCard.DefaultStyle.IMAGE_WIDTH.dp),
--- a/code/messaging/src/main/java/com/adobe/marketing/mobile/aepcomposeui/components/SmallImageCard.kt
+++ b/code/messaging/src/main/java/com/adobe/marketing/mobile/aepcomposeui/components/SmallImageCard.kt
@@ -34,8 +34,6 @@ import com.adobe.marketing.mobile.aepcomposeui.UIEvent
 import com.adobe.marketing.mobile.aepcomposeui.observers.AepUIEventObserver
 import com.adobe.marketing.mobile.aepcomposeui.style.SmallImageUIStyle
 import com.adobe.marketing.mobile.aepcomposeui.utils.UIUtils
-import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.withContext
 
 /**
  * Composable function that renders a small image card UI.
@@ -60,9 +58,7 @@ fun SmallImageCard(
         if (imageUrl.isNullOrBlank()) {
             isLoading = false
         } else {
-            imageBitmap = withContext(Dispatchers.IO) {
-                UIUtils.downloadImage(imageUrl)
-            }
+            imageBitmap = UIUtils.downloadImage(imageUrl)
             isLoading = false
         }
     }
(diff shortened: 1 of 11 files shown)


code/messaging/src/main/java/com/adobe/marketing/mobile/aepcomposeui/components/SmallImageCard.kt
```diff
@@ -50,16 +50,20 @@ fun SmallImageCard(
...
     LaunchedEffect(ui.getTemplate().id) {
         observer?.onEvent(UIEvent.Display(ui))
-        if (imageUrl != null) {
+        if (imageUrl.isNullOrBlank()) {
+            isLoading = false
+        } else {
             imageBitmap = withContext(Dispatchers.IO) {
                 UIUtils.downloadImage(imageUrl)
             }
+            isLoading = false
```
rymorale, line 66: should this be isLoading = true as the image is being downloaded in the else branch?

rymorale (approved):
just had a question regarding the `isLoading` boolean but otherwise the pr LGTM

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/SmallImageTemplateEventHandler.kt
```diff
@@ -32,7 +34,25 @@ internal class SmallImageTemplateEventHandler(private val callback: ContentCardC
...
+
+                // Call the super class to call the callback and send tracking event if needed
+                super.onEvent(event, propositionId)
+
+                // For display event, change the displayed state of the UI to true
+                // after the tracking event is sent for the initial composition
+                if (event is UIEvent.Display) {
+                    smallImageUI.updateState(currentUiState.copy(displayed = true))
+                }
+            }
```
prudrabhat, line 55: - Is there a reason this particular ordering is needed for sending events? Can we not call super for all events?
- Given that this class deals with SmallImageUi only, we can cast it eagerly and bail if it is unrelated to SmallImageUi allowing you to simplify this function

```
  val smallImageUI : SmallImageUI = event.aepUi as? SmallImageUI ?: return
  val currentUiState = smallImageUI.getState() as SmallImageCardUIState
   when (event) {
          is UIEvent.Dismiss -> {
              smallImageUI.updateState(currentUiState.copy(dismissed = true))
          }
          is UIEvent.Display -> {
              smallImageUI.updateState(currentUiState.copy(displayed = true))
          }

          is UIEvent.Interact -> {
              onInteractEvent(event, propositionId)
          }

        super.onEvent(event, propositionId)
    }
}
```

code/messaging/src/main/java/com/adobe/marketing/mobile/aepcomposeui/components/SmallImageCard.kt
prudrabhat, line 66: If the algorithm is  
- `always expect an image and display loading indicator`
- `If image is null dismiss loading indicator, else download image`
- `dismiss loading indicator once image is downloaded`

should `isLoading` be set to false inside the coroutine?

spoorthipujariadobe: `isLoading` has default value of true. Once the coroutine to download the image completes, this line will run and set isLoading to false

code/testapp/src/main/java/com/adobe/marketing/mobile/messagingsample/MessagingApplication.kt
```diff
@@ -22,10 +22,10 @@ import com.adobe.marketing.mobile.edge.identity.Identity
 import com.google.firebase.messaging.FirebaseMessaging
 
 class MessagingApplication : Application() {
-    private val ENVIRONMENT_FILE_ID = "3149c49c3910/4f6b2fbf2986/launch-7d78a5fd1de3-development"
-    private val ASSURANCE_SESSION_ID = ""
+    private val ENVIRONMENT_FILE_ID = "3149c49c3910/473386a6e5b0/launch-6099493a8c97-development"
```
rymorale, line 25: nit: can we store the existing app id in another val? it will save time in the future if we need to go back to different testing.

rymorale: noticed that the staging flag is set to true. is this appid a staging id?

code/messaging/src/main/java/com/adobe/marketing/mobile/aepcomposeui/components/SmallImageCard.kt
rymorale: my bad, i had missed the default value above

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/SmallImageTemplateEventHandler.kt
spoorthipujariadobe: If `displayed = true` is set to true before calling `super.onEvent`, the tracking event, which is only sent when `displayed` is false, will never work. 
Will make the change for casting early 👍

code/testapp/src/main/java/com/adobe/marketing/mobile/messagingsample/MessagingApplication.kt
spoorthipujariadobe: This is prod app id for a different tag property. I accidentally changed it, will revert it back

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/SmallImageTemplateEventHandler.kt
prudrabhat: So calling super before changing the state should work right?

spoorthipujariadobe: Changed to have the super class `onEvent` not be open. Instead, provided an abstract function `handleEvent` to be implemented by individual template listeners. Base class `MessagingEventHandler.onEvent` will always be called first now. It can perform the tasks common to all templates and delegate the specific tasks to sub classes by calling `handleEvent`.

cc: @rymorale

code/messaging/src/main/java/com/adobe/marketing/mobile/aepcomposeui/components/SmallImageCard.kt
spoorthipujariadobe: Changed `UIUtils.downloadImage(url: String)` to use ServiceProvider network service. Since that is already guaranteed to run on a different thread, removed switching content to IO coroutine here

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/MessagingEventHandler.kt
```diff
@@ -21,25 +21,99 @@ import com.adobe.marketing.mobile.services.Log
...
+                    SELF_TAG,
+                    "UI displayed for the first time, sending display tracking event"
+                )
 
-        internal fun track(propositionId: String, interaction: String?, eventType: MessagingEdgeEventType?) {
-            val contentCardSchemaData = ContentCardMapper.instance.getContentCardSchemaData(propositionId)
-            contentCardSchemaData?.track(interaction, eventType)
+                track(propositionId, null, MessagingEdgeEventType.DISPLAY)
+                handleEvent(event, propositionId)
+                callback?.onDisplay(event.aepUi)
```
prudrabhat, line 73: nit: Can we filp the order so that state handling is done before tracking and callbacks.

```
handleEvent(...)
track(..)
callback.onDisplay()
```

code/testapp/src/main/java/com/adobe/marketing/mobile/messagingsample/ScrollingFeedActivity.kt
```diff
@@ -12,58 +12,237 @@
...
-                        propositions.add(proposition)
+
+        // Initialize the ContentCardUIProvider
+        contentCardUIProvider = ContentCardUIProvider(surface)
+
+        // Initialize the ViewModel
+        contentCardViewModel =
+            ViewModelProvider(this, AepContentCardViewModelFactory(contentCardUIProvider)).get(
+                AepContentCardViewModel::class.java
+            )
```
prudrabhat, line 84: This approach should be OK for our testapp, but, I think we should also provide an example implementation where an Activity as a whole uses a view model  (not entirely creating a new one for the sake of the list) with content cards being fetched inside that viewmodel and retrieved as a KotlinFlow / callbackFlow by the UI. We can discuss more tomorrow.

code/messaging/src/main/java/com/adobe/marketing/mobile/aepcomposeui/components/SmallImageCard.kt
spoorthipujariadobe: Thank you for the great explanation! Changed it to use a callback. I'm working on adding tests, facing some weird issues. I'll update the PR in a bit

prudrabhat (approved):

PR #304, opened 2024-11-05
MOB-22249 and MOB-22214 - use rank and expose priority for proposition ordering

sbenedicadb:
## Types of changes

## Checklist:


spoorthipujariadobe (approved):
Looks good, only one question: Why does getPriority and getRank return `Integer` instead of primitive type `int`? We never return a null value for either, wondering what benefit the `Integer` class type has

rymorale (approved):
LGTM, same observation as @spoorthipujariadobe regarding the use of `Integer` vs `int`. let's use `int` if possible.

PR #318, opened 2024-12-20
[MOB-22579] Added custom interaction for card click

navratan-soni:
## Description
Currently in Android Content cards, there is no way to identify if the card is clicked based on interaction ID.
If user need to identify the same in the callback they need to do so in else block
Like this : 
Screenshot 2024-12-21 at 12 13 17 AM

But after this change, the way to handle the callback click will be like this
Screenshot 2024-12-21 at 12 13 03 AM

This also bring it to the parity with iOS handling the interaction same way
iOS reference : 
https://github.com/adobe/aepsdk-messaging-ios/blob/main/AEPMessaging/Sources/UI/UIConstants.swift#L59
https://github.com/adobe/aepsdk-messaging-ios/blob/973ef4ba35b86d1a28782a3dac5c9c1cfc28c703/AEPMessaging/Sources/UI/ContentCards/Template/BaseTemplate.swift#L63

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:


rymorale (approved):

code/messaging/src/main/java/com/adobe/marketing/mobile/aepcomposeui/components/SmallImageCard.kt
```diff
@@ -74,7 +74,7 @@ fun SmallImageCard(
     AepCardComposable(
         cardStyle = style.cardStyle,
         onClick = {
-            observer?.onEvent(UIEvent.Interact(ui, UIAction.Click(ui.getTemplate().id, ui.getTemplate().actionUrl)))
+            observer?.onEvent(UIEvent.Interact(ui, UIAction.Click(AepUIConstants.InteractionID.CARD_CLICKED, ui.getTemplate().actionUrl)))
```
spoorthipujariadobe, line 77: We already had the content card id added to the click event. Why are we changing this to a general `Card clicked` event? How would the customer differentiate which card was clicked, when multiple cards are displayed?

PR #323, opened 2025-01-13
Fix content card image caching, bump version to 3.3.0-beta4

rymorale:
## Description
- Compress the bitmap as PNG prior to caching to resolve cache retrieval issues.
- add more logging around image caching

## Types of changes

## Checklist:

### DIFF
--- a/code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardImageManager.kt
+++ b/code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardImageManager.kt
@@ -72,7 +72,7 @@ object ContentCardImageManager {
         // Convert the InputStream to a Bitmap
         if (inputStream != null) {
             try {
-                val bitmap = BitmapFactory.decodeStream(inputStream, null, BitmapFactory.Options())
+                val bitmap = BitmapFactory.decodeStream(inputStream)
                 if (bitmap == null) {
                     Log.warning(
                         MessagingConstants.LOG_TAG,
--- a/code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardImageManager.kt
+++ b/code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardImageManager.kt
@@ -22,8 +22,8 @@ import com.adobe.marketing.mobile.services.caching.CacheEntry
 import com.adobe.marketing.mobile.services.caching.CacheExpiry
 import com.adobe.marketing.mobile.services.caching.CacheResult
 import com.adobe.marketing.mobile.services.caching.CacheService
+import java.io.ByteArrayOutputStream
 import java.io.InputStream
-import java.nio.ByteBuffer
 
 object ContentCardImageManager {
     private val SELF_TAG: String = "ContentCardManager"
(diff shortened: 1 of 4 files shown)


code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardImageManager.kt
```diff
@@ -22,8 +22,8 @@ import com.adobe.marketing.mobile.services.caching.CacheEntry
 import com.adobe.marketing.mobile.services.caching.CacheExpiry
 import com.adobe.marketing.mobile.services.caching.CacheResult
 import com.adobe.marketing.mobile.services.caching.CacheService
+import java.io.ByteArrayOutputStream
```
prudrabhat, line 25: few comments un-related to your change while going through some of this class. If possible, would you be able to accommodate them here? These should be small, but if you think the scope becomes too large, feel free to add them as a follow up, but before we release the next snapshot.

- Is this utility being used outside the MessagingExtension? Can this and its methods be `internal`?
- Since this is an `object` and there is a possibility of it being initialized even before context (needed by cache service and device info service) is set, can you please use the cache service on demand where required.
- `defaultCacheName` just mirrors the publicly available constant. Can we directly use `MessagingConstants.CONTENT_CARD_CACHE_SUBDIRECTORY` when necessary.
- Looking at `downloadAndCacheImageBitmap` it appears that the _caching_ result is unrelated to the completion result sent back to the caller - i.e cache failure would also result in `completion(Result.success(bitmap))`. 
Given that we are increasingly interested in reducing the load time of images, can we call the completion before we cache? That way the caller does not really need to wait for the caching to complete in order to display the image at present.

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardImageManager.kt
```diff
@@ -72,7 +72,22 @@ object ContentCardImageManager {
         // Convert the InputStream to a Bitmap
         if (inputStream != null) {
             try {
-                completion(Result.success(BitmapFactory.decodeStream(inputStream)))
+                val bitmap = BitmapFactory.decodeStream(inputStream)
```
prudrabhat, line 75: - Can we flip the if clause to `if (inputStream == null) {` here and return early with completion failure for a clearer flow.
- We should also close the `inputstream` here after reading the content as a bitmap.

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardImageManager.kt
```diff
@@ -133,11 +148,9 @@ object ContentCardImageManager {
...
         try {
             val imageInputStream: InputStream = imageBitmap.let { bitmap ->
-                val byteArray = ByteArray(bitmap.byteCount)
-                val buffer = ByteBuffer.wrap(byteArray)
-                bitmap.copyPixelsToBuffer(buffer)
-                buffer.rewind() // Reset the buffer position to the beginning
-                byteArray.inputStream() // Create InputStream from byte array
+                val byteArrayOutputStream = ByteArrayOutputStream()
+                bitmap.compress(Bitmap.CompressFormat.PNG, 100, byteArrayOutputStream)
+                byteArrayOutputStream.toByteArray().inputStream()
```
prudrabhat, line 153: We will need to close the stream here - use `use` for automatic closure
```
byteArrayOutputStream.use { baos ->
                baos.toByteArray().inputStream()
            }
```

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardImageManager.kt
rymorale: l'll get these fixed up in this pr.

prudrabhat (approved):
LGTM. Let's add tests to cover the missing lines in `getImageBitmapFromCache` as they are the crux of this change.

PR #329, opened 2025-01-24
Fix click behavior for button and image in SmallImageCard

spoorthipujariadobe:
## Description

1. Added a new field `interactionSource` to `AepButtonStyle` to allow customizing the appearance / behavior of the button in different states.
2. Removed explicit `onClick` fields in `AepImageComposable` and `AepIconComposable`. It can be added/overridden using Modifier. This ensures that image in the SmallImageCard inherits the parent card click behavior by default.

## Related Issue

 [MOB-22770](https://jira.corp.adobe.com/browse/MOB-22770)
 [MOB-22771](https://jira.corp.adobe.com/browse/MOB-22771)

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [X] I have signed the [Adobe Open Source CLA](http://opensource.adobe.com/cla.html).
- [X] My code follows the code style of this project.
- [X] I have read the **CONTRIBUTING** document.
- [X] I have added tests to cover my changes.
- [X] All new and existing tests passed.

### DIFF
--- a/code/messaging/src/main/java/com/adobe/marketing/mobile/aepcomposeui/components/SmallImageCard.kt
+++ b/code/messaging/src/main/java/com/adobe/marketing/mobile/aepcomposeui/components/SmallImageCard.kt
@@ -12,6 +12,7 @@
 package com.adobe.marketing.mobile.aepcomposeui.components
 
 import android.graphics.Bitmap
+import androidx.compose.foundation.clickable
 import androidx.compose.foundation.isSystemInDarkTheme
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.size
@@ -84,10 +85,11 @@ fun SmallImageCard(
                     drawableId = it.drawableId,
                     // todo check if we can remember this calculation so that it is not repeated for recompositions
                     iconStyle = style.dismissButtonStyle.apply {
-                        modifier = (modifier ?: Modifier).align(style.dismissButtonAlignment)
-                    },
-                    onClick = {
-                        observer?.onEvent(UIEvent.Dismiss(ui))
+                        modifier = (modifier ?: Modifier)
+                            .align(style.dismissButtonAlignment)
+                            .clickable {
+                                observer?.onEvent(UIEvent.Dismiss(ui))
+                            }
                     }
                 )
             }
--- a/code/messaging/src/main/java/com/adobe/marketing/mobile/aepcomposeui/components/AepButtonComposable.kt
+++ b/code/messaging/src/main/java/com/adobe/marketing/mobile/aepcomposeui/components/AepButtonComposable.kt
@@ -11,9 +11,11 @@
 
 package com.adobe.marketing.mobile.aepcomposeui.components
 
+import androidx.compose.foundation.interaction.MutableInteractionSource
 import androidx.compose.material3.Button
 import androidx.compose.material3.ButtonDefaults
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.remember
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.tooling.preview.Preview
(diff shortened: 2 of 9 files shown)


rymorale (approved):

code/messaging/src/main/java/com/adobe/marketing/mobile/aepcomposeui/components/SmallImageCard.kt
```diff
@@ -84,10 +85,11 @@ fun SmallImageCard(
...
                     // todo check if we can remember this calculation so that it is not repeated for recompositions
                     iconStyle = style.dismissButtonStyle.apply {
-                        modifier = (modifier ?: Modifier).align(style.dismissButtonAlignment)
-                    },
-                    onClick = {
-                        observer?.onEvent(UIEvent.Dismiss(ui))
+                        modifier = (modifier ?: Modifier)
+                            .align(style.dismissButtonAlignment)
+                            .clickable {
+                                observer?.onEvent(UIEvent.Dismiss(ui))
```
prudrabhat, line 91: I remember that `clickable`'s on the modifier cannot be accumulated (defining 5 `clickable`'s will only trigger the final one in the hierarchy...). Were you able to see a clickable passed from the app(if defined) and the one defined here both fire? Or am I not understanding this right... 🤔

spoorthipujariadobe: That's correct, only the final `clickable` in the hierarchy is called. Since the SDK needs to send out tracking events on dismiss click (and hide the card when we have a container), we want our `clickable` to be called over any implementation that the app developer might added using `SmallImageUIStyle.Builder().dismissButtonStyle(AepIconStyle(modifier = Modifier .clickable {}))`. Which is why I've added this right at the end here after the default and custom styles as been merged to ensure it gets precedence.

prudrabhat: Ah, I see. Got it! This makes sense 👍🏼 . 
Though we can  this particular change does not alter the in public api, it would be good to have an idea about how we want to extend this use-case when an app would be allowed to override an action on an icon (like a custom one instead of dismiss) and its impact on the current public api.

prudrabhat (approved):

code/messaging/src/main/java/com/adobe/marketing/mobile/aepcomposeui/components/SmallImageCard.kt
spoorthipujariadobe: That's a good point! Material Design components like [Button](https://developer.android.com/reference/kotlin/androidx/compose/material3/package-summary#Button(kotlin.Function0,androidx.compose.ui.Modifier,kotlin.Boolean,androidx.compose.ui.graphics.Shape,androidx.compose.material3.ButtonColors,androidx.compose.material3.ButtonElevation,androidx.compose.foundation.BorderStroke,androidx.compose.foundation.layout.PaddingValues,androidx.compose.foundation.interaction.MutableInteractionSource,kotlin.Function1)) and [Card](https://developer.android.com/reference/kotlin/androidx/compose/material3/package-summary#Card(kotlin.Function0,androidx.compose.ui.Modifier,kotlin.Boolean,androidx.compose.ui.graphics.Shape,androidx.compose.material3.CardColors,androidx.compose.material3.CardElevation,androidx.compose.foundation.BorderStroke,androidx.compose.foundation.interaction.MutableInteractionSource,kotlin.Function1) do this by exposing a separate `onClick` field. This is something we can add to our style classes for the elements we support, example `AepIconStyle` can have additional `onClick` which can be customized by the app developer and we use that value when merging modifiers.

PR #330, opened 2025-01-29
[MOB-22211] Add overloaded updatePropositions API that allows for completion. Push to in-app implementation complete and manually tested.

sbenedicadb:
## Description

- [MOB-22211] Add overloaded updatePropositions API that allows for completion. Push to in-app implementation complete and manually tested.

## Types of changes

## Checklist:

### DIFF
--- a/code/messaging/src/main/java/com/adobe/marketing/mobile/MessagingPushPayload.java
+++ b/code/messaging/src/main/java/com/adobe/marketing/mobile/MessagingPushPayload.java
@@ -99,21 +99,24 @@ public class MessagingPushPayload {
     /**
      * Contains push payload keys to be added to Intent.extras via {@link #putDataInExtras(Intent)}
      */
-    private static final List<String> _pushPayloadKeys = new ArrayList<String>() {{
-        add(MessagingConstants.Push.PayloadKeys.TITLE);
-        add(MessagingConstants.Push.PayloadKeys.BODY);
-        add(MessagingConstants.Push.PayloadKeys.SOUND);
-        add(MessagingConstants.Push.PayloadKeys.BADGE_NUMBER);
-        add(MessagingConstants.Push.PayloadKeys.NOTIFICATION_VISIBILITY);
-        add(MessagingConstants.Push.PayloadKeys.NOTIFICATION_PRIORITY);
-        add(MessagingConstants.Push.PayloadKeys.CHANNEL_ID);
-        add(MessagingConstants.Push.PayloadKeys.ICON);
-        add(MessagingConstants.Push.PayloadKeys.IMAGE_URL);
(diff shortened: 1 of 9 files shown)


code/messaging/src/main/java/com/adobe/marketing/mobile/MessagingPushPayload.java
```diff
@@ -95,6 +96,28 @@ static final class ActionButtons {
                 }
             };
 
+    /**
+     * Contains push payload keys to be added to Intent.extras via {@link #putDataInExtras(Intent)}
+     */
+    private static final List<String> _pushPayloadKeys =
```
spoorthipujariadobe, line 102: nit: It's not common to have `_` at the beginning of a variable name. Is it ok if I remove it?

code/messaging/src/phone/java/com/adobe/marketing/mobile/Messaging.java
```diff
@@ -390,12 +447,26 @@ public void call(final Event event) {
...
+     *
+     * @param surfaces A {@code List<Surface>} containing {@link Surface}s to be used for retrieving
+     *     propositions
+     * @param callback An optional callback to be called once the proposition response has been
+     *     processed by the Messaging extension
+     */
+    public static void updatePropositionsForSurfaces(
+            @NonNull final List<Surface> surfaces,
+            @Nullable final AdobeCallback<Boolean> callback) {
+        if (surfaces.isEmpty()) {
```
spoorthipujariadobe, line 465: nit: In Java the @NonNull annotation is not really enforced. The compiler will give a warning if this API is called with a null value but the build will not fail. To cover our bases, is it ok if I add back the null check?

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/MessagingExtension.java
```diff
@@ -46,6 +48,9 @@ public final class MessagingExtension extends Extension {
     final ContentCardRulesEngine contentCardRulesEngine;
     private SerialWorkDispatcher<Event> serialWorkDispatcher;
 
+    private static final Object completionHandlersMutex = new Object();
+    private static List<CompletionHandler> completionHandlers = new ArrayList<>();
```
spoorthipujariadobe, line 52: Wondering if it would be better for performance to use `ConcurrentLinkedQueue` here instead of a regular ArrayList with `synchronized` blocked for reads and writes. Thoughts?

spoorthipujariadobe: Discussed offline. Since this API only gets called when the push notification is clicked, we don't expect it to get be called frequently. Leaving this as is to revisit if we face performance issues.

spoorthipujariadobe (approved):

PR #349, opened 2025-02-27
MOB-22412: reduce push token sync frequency

rymorale:
## Description
Update push token sync network request to be sent only when the values change. A new boolean configuration key value pair has been added `messaging.pushForceSync` which can be used to override this new behavior.

## Types of changes

## Checklist:


code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/MessagingConstants.java
```diff
@@ -31,7 +31,7 @@ public final class MessagingConstants {
     static final int DEFAULT_TIMEOUT = 5;
     static final long RESPONSE_CALLBACK_TIMEOUT = 10000; // 10 seconds
     static final long CACHE_EXPIRY_TIME = 604800000; // 7 days in milliseconds
-    static final int HALF_SECOND_IN_MILLIS = 500;
+    static final String DATA_STORE_NAME = "AdobeMobile_Messaging";
```
kevinlind, line 34: It's common to use `DATA_STORE_NAME = EXTENSION_NAME`, unless you have multiple data stores.

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/MessagingExtension.java
```diff
@@ -882,6 +898,19 @@ private Map<String, Object> getXDMSharedState(final String extensionName, final
     private boolean eventIsValid(final Event event) {
         return event != null && event.getEventData() != null;
     }
+
+    private void addPushTokenToSharedState(final String pushToken, final Event event) {
```
kevinlind, line 902: It might be me but `add` here makes me think the push token is being added to an existing shared state. Can you rename to something such as `sharePushToken` or even `createSharedState`?

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/InternalMessagingUtils.java
```diff
@@ -560,6 +562,83 @@ static String getSharedStateEcid(final Map<String, Object> edgeIdentityState) {
...
+     * service.
+     *
+     * @return The Messaging extension {@link NamedCollection}.
+     */
+    private static NamedCollection getNamedCollection() {
+        final DataStoring dataStoreService = ServiceProvider.getInstance().getDataStoreService();
+        return dataStoreService.getNamedCollection(MessagingConstants.DATA_STORE_NAME);
+    }
+
+    @Nullable static String getExistingPushToken() {
```
kevinlind, line 636: Rename to `getPushTokenFromPersistence` to make it clear this is a stored value.

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/MessagingExtension.java
```diff
@@ -468,10 +487,7 @@ void handlePushToken(final Event event) {
         }
 
         // Update the push token to the shared state
-        final HashMap<String, Object> messagingSharedState = new HashMap<>();
-        messagingSharedState.put(
-                MessagingConstants.SharedState.Messaging.PUSH_IDENTIFIER, pushToken);
-        getApi().createSharedState(messagingSharedState, event);
+        addPushTokenToSharedState(pushToken, event);
```
kevinlind, line 490: You'll need to set the token in persistence here.

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/MessagingExtension.java
```diff
@@ -882,6 +898,19 @@ private Map<String, Object> getXDMSharedState(final String extensionName, final
...
         return event != null && event.getEventData() != null;
     }
+
+    private void addPushTokenToSharedState(final String pushToken, final Event event) {
+        if (StringUtils.isNullOrEmpty(pushToken)) {
+            Log.debug(
+                    MessagingConstants.LOG_TAG,
+                    SELF_TAG,
+                    "Push token is null or empty. Not adding the push token to the shared state.");
+            return;
```
kevinlind, line 908: You'll still need to set the shared state if the push token is null or empty. Change this logic to simply not add the push token to the `newMessagingState`. The reason is if the push token is set, then removed, the shared state needs to be updated with a cleared push token to reflect the change in state.

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/InternalMessagingUtils.java
```diff
@@ -560,6 +562,83 @@ static String getSharedStateEcid(final Map<String, Object> edgeIdentityState) {
...
+     */
+    static boolean shouldSyncPushToken(
+            @Nullable final Map<String, Object> configSharedState,
+            @Nullable final String newPushToken) {
+        if (StringUtils.isNullOrEmpty(newPushToken)) {
+            Log.debug(
+                    MessagingConstants.LOG_TAG,
+                    "shouldSyncPushToken",
+                    "New push token is null or empty, push token will not be synced.");
+            return false;
```
kevinlind, line 581: In Messaging, if the new push token is null/empty while the persisted token is not, do you need to sync the empty push token to Profile?

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/InternalMessagingUtils.java
```diff
@@ -560,6 +562,83 @@ static String getSharedStateEcid(final Map<String, Object> edgeIdentityState) {
...
+                        MessagingConstants.SharedState.Configuration.PUSH_REGISTRATION_PAUSED,
+                        false);
+
+        if (isPushRegistrationPaused) {
+            Log.debug(
+                    MessagingConstants.LOG_TAG,
+                    "shouldSyncPushToken",
+                    "Push registration is paused, push token will not be synced.");
+            return false;
+        }
```
kevinlind, line 596: The logic we want is not to simply pause all push requests, but to either sync on all requests or to sync only when the value changes.
I'd suggest

1. Change config flag to `messaging.pushForceSync`
2. Default for pushForceSync is false, meaning only sync on value change.
3. Update check on line 599 to `existingPushToken != null && existingPushToken.equals(newPushToken) && !pushForceSync)`

kevinlind (commented):
Additionally, if we are syncing the push token but the request is within a second of the previous sync then the request should be ignored. Are you planning on adding that logic here or in another PR?

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/MessagingExtension.java
rymorale: the token is added to persistence prior to this method being called in InternalMessagingUtils:
```java
 // persist the push token in the messaging named collection
        final NamedCollection messagingNamedCollection = getNamedCollection();
        messagingNamedCollection.setString(
                MessagingConstants.SharedState.Messaging.PUSH_IDENTIFIER, newPushToken);
```

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/InternalMessagingUtils.java
rymorale: good callout, i'll still sync a token if it is empty.

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/InternalMessagingUtils.java
rymorale: i'll get this fixed

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/InternalMessagingUtils.java
rymorale: we ignore push sync events which do not have a valid push token on android and iOS. this is the existing behavior but i think it would be good to discuss if an empty token should still be synced.

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/MessagingExtension.java
rymorale: this null or empty check shouldn't be needed as push sync requests containing null or empty tokens are discarded. i'll remove the check.

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/InternalMessagingUtils.java
kevinlind: Yes, we can discuss if null tokens are allowed and will clear the local token. If we decide to make that change we can take it up in another PR.

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/MessagingExtension.java
kevinlind: The null check is actually still needed because this API is called on registration and is passed the persisted push token, which may be null. Let's leave the check in.

kevinlind (approved):

PR #362, opened 2025-05-19
Handle reset identities event by clearing the push token from messaging shared state and persistence

rymorale:
## Description
A new ECID is generated when the resetIdentities API is called. The Messaging extension should clear the existing push token from Messaging shared state and persistence when a resetIdentities event is received. The app developer will then be responsible for syncing the push token with the newly generated ECID.

## Types of changes

## Checklist:

### DIFF
--- a/code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/InternalMessagingUtils.java
+++ b/code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/InternalMessagingUtils.java
@@ -186,22 +186,6 @@ class InternalMessagingUtils {
     }
 
     /**
-     * Determines if the passed in {@code Event} is an edge identity reset complete event.
-     *
-     * @param event An Edge Identity Reset Complete {@link Event}.
-     * @return {@code boolean} indicating if the passed in event is an edge identity reset complete
-     *     event.
-     */
-    static boolean isEdgeIdentityResetComplete(final Event event) {
-        if (event == null) {
-            return false;
-        }
-
-        return EventType.EDGE_IDENTITY.equalsIgnoreCase(event.getType())
-                && EventSource.RESET_COMPLETE.equalsIgnoreCase(event.getSource());
-    }
-
-    /**
      * Determines if the passed in {@code Event} is a messaging request content event.
      *
      * @param event A Messaging Request Content {@link Event}.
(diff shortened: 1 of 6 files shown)


code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/InternalMessagingUtils.java
```diff
@@ -647,6 +661,16 @@ private static boolean isPushTokenSyncTimeoutExpired(final long eventTimestamp)
...
     // ========================================================================================
+    /**
+     * Persists the provided push token in the Messaging extension {@link NamedCollection}.
+     *
+     * @param pushToken A {@code String} containing the push token to be persisted
+     */
+    static void persistPushToken(final String pushToken) {
+        final NamedCollection messagingNamedCollection = getNamedCollection();
+        messagingNamedCollection.setString(
+                MessagingConstants.NamedCollectionKeys.Messaging.PUSH_IDENTIFIER, pushToken);
```
kevinlind, line 672: nit: It is safer to check if `pushToken` is null and if so call `NamedCollection.remove()` instead. It is true that the underlying `SharedPreferencesEditor.putString` will remove the value if passed null, but `NamedCollection.setString` doesn't explicitly state that behavior. My concern is if the `NamedCollection` implementation is changed then passing null may no longer remove the value.

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/MessagingExtension.java
```diff
@@ -897,17 +929,14 @@ private Map<String, Object> getXDMSharedState(final String extensionName, final
...
     private void createMessagingSharedState(final String pushToken, final Event event) {
-        if (StringUtils.isNullOrEmpty(pushToken)) {
-            Log.debug(
-                    MessagingConstants.LOG_TAG,
-                    SELF_TAG,
-                    "Push token is null or empty. Not adding the push token to the shared state.");
-            return;
-        }
         final Map<String, Object> newMessagingState = new HashMap<>();
         newMessagingState.put(MessagingConstants.SharedState.Messaging.PUSH_IDENTIFIER, pushToken);
```
kevinlind, line 941: Check if `pushToken` is not null or empty before adding to `newMessagingState`. When the push token is cleared, then the Messaging shared state will simply be empty.

kevinlind (approved):

PR #391, opened 2025-09-15
MOB-23986 : Adding Notification Delete Listener

sagar-sharma-adobe:
## Types of changes

## Checklist:

- [X] I have signed the [Adobe Open Source CLA](http://opensource.adobe.com/cla.html).
- [X] My code follows the code style of this project.
- [X] I have read the **CONTRIBUTING** document.
- [X] All new and existing tests passed.

### DIFF
--- a/code/messaging/src/test/java/com/adobe/marketing/mobile/messaging/NotificationInteractionReceiverTest.kt
+++ b/code/messaging/src/test/java/com/adobe/marketing/mobile/messaging/NotificationInteractionReceiverTest.kt
@@ -19,13 +19,13 @@ import org.mockito.Mockito.mock
 import org.mockito.Mockito.mockStatic
 import org.mockito.Mockito.never
 
-class NotificationDeleteReceiverTest {
+class NotificationInteractionReceiverTest {
 
     @Test
     fun `onReceive should call Messaging handleNotificationResponse with dismiss action`() {
         val context = mock(Context::class.java)
         val intent = mock(Intent::class.java)
-        val receiver = NotificationDeleteReceiver()
+        val receiver = NotificationInteractionReceiver()
 
         // Mock static method
         mockStatic(Messaging::class.java).use { messagingMock ->
@@ -39,7 +39,7 @@ class NotificationDeleteReceiverTest {
     @Test
     fun `onReceive with null intent should not call Messaging handleNotificationResponse`() {
         val context = mock(Context::class.java)
-        val receiver = NotificationDeleteReceiver()
+        val receiver = NotificationInteractionReceiver()
         val intent = mock(Intent::class.java)
 
         mockStatic(Messaging::class.java).use { messagingMock ->


code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/NotificationDeleteReceiver.kt
```diff
@@ -0,0 +1,30 @@
...
+*/
+
+package com.adobe.marketing.mobile.messaging
+
+import android.content.BroadcastReceiver
+import android.content.Context
+import android.content.Intent
+import com.adobe.marketing.mobile.Messaging
+
+class NotificationDeleteReceiver : BroadcastReceiver() {
```
spoorthipujariadobe, line 19: Can we also declare this in the manifest file?

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/NotificationDeleteReceiver.kt
```diff
@@ -0,0 +1,30 @@
...
+    private val dismissAction: String = "Dismiss"
```
siddique-adobe, line 20: Instead of hardcoding the action name here, can't we extract it from the intent data?, So that the same code can be used for further use cases

code/messaging/src/test/java/com/adobe/marketing/mobile/messaging/NotificationDeleteReceiverTest.kt
```diff
@@ -0,0 +1,53 @@
...
+    fun `onReceive should call Messaging handleNotificationResponse with dismiss action`() {
+        val context = mock(Context::class.java)
+        val intent = mock(Intent::class.java)
+        val receiver = NotificationDeleteReceiver()
+
+        // Mock static method
+        mockStatic(Messaging::class.java).use { messagingMock ->
+            receiver.onReceive(context, intent)
+            messagingMock.verify {
+                Messaging.handleNotificationResponse(intent, false, "Dismiss")
```
siddique-adobe, line 34: Can we add the test for actions name verification so that the value shouldn't be changed during the broadcast or within the broadcast

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/NotificationDeleteReceiver.kt
sagar-sharma-adobe: Yes, it is already added in the manifest

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/NotificationDeleteReceiver.kt
sagar-sharma-adobe: Good point, moved to the constant file

code/messaging/src/test/java/com/adobe/marketing/mobile/messaging/NotificationDeleteReceiverTest.kt
sagar-sharma-adobe: Yes, we will refer to the constant file and make sure both values alwaysa match.

PR #419, opened 2026-02-13
Handle inbox schema type propositions + inbox display tracking

spoorthipujariadobe:
## Description

The tracking implementation provides automatic display tracking for inbox propositions while continuing to maintain the separation between UI components (`aepcomposeui`) and messaging SDK logic (`messaging`).

#### 1. `InboxUIState.kt`
- `InboxUIState.Success` contains a `displayed` flag.
- Each new state instance starts with `displayed = false`
- Prevents duplicate tracking on recomposition while allowing tracking for new inbox instances.

#### 2. `InboxEvent.kt`
- New sealed class for inbox container-level events (display, sort, filter, etc.)
- Currently implements `InboxEvent.Display`

#### 3. `AepInboxEventObserver` / `InboxEventObserver`
- New interface extending `AepUIEventObserver` to add inbox-level event handling
- `InboxEventObserver` - New implementation that coordinates both inbox and item-level events
- Accepts `vararg` item observers (e.g., `ContentCardEventObserver`) for extensibility and delegates item event handling to these observers.
- Allows content provider to set a callback for inbox events and delegates inbox event handling to the content provider. 

#### 4. `ContentCardMapper.kt`
- Extended to store inbox proposition items in addition to content cards
- Inbox propositions stored during state creation in `MessagingInboxProvider` and retrieved during display tracking in `InboxEventObserver`.

#### 5. `AepInbox.kt`
- Updated to accept `AepInboxEventObserver` (non-optional)
- Uses `LaunchedEffect(uiState.template.id, uiState.displayed)` to trigger display tracking once per composition

#### **Data Flow**

1. `MessagingInboxProvider` accepts `InboxEventObserver` and sets the callback that's called on inbox events.
2. `MessagingInboxProvider` stores inbox `PropositionItem` in `ContentCardMapper` and creates state.
4. Returns `InboxUIState.Success` (with `displayed = false`)
5. `AepInbox` composable displays and checks `displayed` flag
6. Calls `observer.onInboxEvent(InboxEvent.Display)`
7. `InboxEventObserver` retrieves `PropositionItem` from `ContentCardMapper` and tracks display event via Messaging SDK
9. `InboxEventObserver` calls `updateInboxState` of the content provider with new success state where `displayed=true`, which in turn updates the state flow.

## Types of changes

- [x] New feature (non-breaking change which adds functionality)

## Checklist:

- [x] I have signed the [Adobe Open Source CLA](http://opensource.adobe.com/cla.html).
- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.


code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/InboxEventObserver.kt
```diff
@@ -0,0 +1,71 @@
...
+        when (event) {
+            is InboxEvent.Display -> {
+                val uiState = event.inboxUIState
+                if (!uiState.displayed) {
+                    ContentCardMapper.instance.getInboxPropositionItem(event.inboxUIState.template.id)
+                        ?.track(MessagingEdgeEventType.DISPLAY)
+                    uiState.displayed = true
+                }
+            }
+        }
```
prudrabhat: While this inbox tracking approach (`uiState.displayed = true` inside observer) works, it’s more implicit and couples tracking logic to mutable state living inside the event payload.
Would you be willing to consider the pattern we follow for the cards i.e derive new state (`uiState.copy(displayed = true`) and publish it via a single owner.

Something to note is that `AepUI` cards work because each card is an object with internal mutable state (BaseAepUI + updateState), so MessagingEventHandler can safely do: 
1. read event.aepUi.getState()
2. compute copy
3. call event.aepUi.updateState(...)

For inbox, there is no equivalent InboxUI object with updateState(...) because state comes from `MessagingInboxProvider` as a `Flow<InboxUIState>`. So I am thinking the natural single owner is the `MessagingInboxProvider`.

Do you think this can work?

spoorthipujariadobe: Is the suggestion to have `MessagingInboxProvider` own the changes to `displayed` state? It is called by the app developer before the actual display. They might choose to call `getInboxUI` but never pass the `Flow<InboxUIState>` to `AepInbox` composable. So the state needs to be flipped in the `LaunchedEffect` of `AepInbox`. But right now, the composable in the UI layer is completely decoupled from the data provider in the Messaging layer so passing the displayed signal from `AepInbox` to `MessagingInboxProvider` is not possible.
Thinking more about this, we might not even need the `displayed` flag if the `LaunchedEffect` in `AepInbox` uses the template id i.e activity id as the key since it remains the same across network calls. How about I do away with the flag and its mutation completely?

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardMapper.kt
```diff
@@ -61,8 +62,36 @@ class ContentCardMapper private constructor() {
...
+     *
+     * @param inboxId the inbox activity ID (from InboxTemplate.id)
+     * @return the inbox [PropositionItem], or null if not stored
+     */
+    fun getInboxPropositionItem(inboxId: String): PropositionItem? {
+        if (inboxId.isEmpty()) {
+            return null
+        }
+        return inboxPropositionItemMap[inboxId]
+    }
```
prudrabhat: Is this public for custom app Inbox implementations?

code/messaging/src/main/java/com/adobe/marketing/mobile/aepcomposeui/components/AepInbox.kt
```diff
@@ -53,6 +55,11 @@ fun AepInbox(
         }
 
         is InboxUIState.Success -> {
+            LaunchedEffect(uiState.template.id, uiState.displayed) {
+                if (!uiState.displayed) {
+                    observer.onInboxEvent(InboxEvent.Display(uiState))
+                }
+            }
```
prudrabhat, line 62: Do we need to gate this based on displayed anymore? Wondering if the tracking logic can be based on the displayed check..

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/InboxEventObserver.kt
```diff
@@ -0,0 +1,33 @@
...
+     * Performs tracking for inbox events.
+     */
+    internal fun onInboxEvent(event: InboxEvent) {
+        when (event) {
+            is InboxEvent.Display -> {
+                ContentCardMapper.instance.getInboxPropositionItem(event.inboxUIState.template.id)
+                    ?.track(MessagingEdgeEventType.DISPLAY)
+            }
+        }
+    }
```
prudrabhat: Can this be a lambda inside `MessagingInboxEventObserver` directly? That way you can rename `MessagingInboxEventObserver` to `InboxEventObserver` to match `ContentCardEventObserver`?

Even simpler would be to make this tracking observer implement `AepInboxEventObserver` and implement `onInboxEvent`.

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardMapper.kt
spoorthipujariadobe: It is used by the observer to call the tracking APIs. I will make it internal, doesn't need to be public

code/messaging/src/main/java/com/adobe/marketing/mobile/aepcomposeui/components/AepInbox.kt
spoorthipujariadobe: We're using `uiState.displayed ` as a key in the LaunchedEffect to ensure onInboxEvent is triggered when refresh resets state to `displayed=false`.  Without the `!uiState.displayed` check, the LaunchedEffect would also get triggered when displayed is changed from `false` to `true`, leading to an extra, unwanted onInboxEvent callback.

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ContentCardUIProvider.kt
```diff
@@ -82,7 +100,20 @@ class ContentCardUIProvider(val surface: Surface) : AepUIContentProvider {
      * so this method is only needed for manual refresh operations.
      */
     override suspend fun refreshContent() {
-        _contentFlow.update { fetchContent() }
+        val templateResult = fetchContent()
+        _contentFlow.value = templateResult
```
prudrabhat: nit: lets use `_contentFlow.update { ...)` when updating values.

prudrabhat (approved):
Looks good. Just one suggestion!

PR #436, opened 2026-05-24
Add pushTracking.receive event for push delivery tracking

akhiljain1907:
## Summary

Adds a third AJO push interaction event — `pushTracking.receive` — which records
that a push notification was successfully delivered to the device.

The event is emitted through a new public API:

- **Automatic display & tracking:** `MessagingService.handleRemoteMessage()` automatically dispatches the receive event after posting the notification.
- **Customer-handled display:** Apps that process and display notifications themselves must explicitly call `Messaging.trackPushReceived(remoteMessage)` from their `FirebaseMessagingService#onMessageReceived()` implementation to record the receive event. Customer should initialize there SDK on Application onCreate method and make sure when they call trackPushReceived api, SDK is initialized.

Also adds an opt-in-by-default **self-init** path in `MessagingService.handleRemoteMessage`
that bootstraps the SDK from a cached `appId` when the host app's foreground
initialization hasn't completed yet (killed-state cold-start race).

No integration changes are required for customers using the documented `MessagingService`
push handling flow. Customers implementing their own notification display logic can
optionally adopt `Messaging.trackPushReceived(...)` to enable receive tracking.

## Background

Today AJO tracks push *open* (`pushTracking.applicationOpened`) and *custom
action* (`pushTracking.customAction`). A delivered-but-not-interacted-with
push produces no Experience Event, so AJO's "delivered" metric currently has
no device-side signal — it's based on FCM's send-side acknowledgement only.

## What changed

| File | Change |
|---|---|
| `Messaging.java` | New public `trackPushReceived(remoteMessage)` API that dispatches a `pushTracking.receive` event to Edge. Intended for customer-managed notification display flows where `MessagingService` is not used. |
| `MessagingService.java` | Automatic notification handling now dispatches `pushTracking.receive` via `Messaging.trackPushReceived(...)` after the notification is posted. Replaced the legacy dead-letter `EventSource.RESPONSE_CONTENT` dispatch (no listener consumed it). Added `selfInit(Context, Runnable onInitComplete)` that reads the cached `appId` from `AdobeMobile_ConfigState/config.appID` and calls `MobileCore.initialize(application, InitOptions, callback)`. When self-init is engaged, the receive-event dispatch is deferred into the `MobileCore.initialize` completion callback — the FCM service thread is never blocked, and the dispatch is guaranteed to run after extensions are registered. |
| `MessagingExtension.java` | New routing branch in `processEvent()` that matches `isPushNotificationReceivedEvent` and routes to the existing `handleTrackingInfo` machinery — reuses the same code path as `applicationOpened` / `customAction`. 
| `InternalMessagingUtils.java` | New `isPushNotificationReceivedEvent(Event)` detector. Pattern matches existing `isSchemaConsequence`, `isTrackingPropositionsEvent`, etc. |
| `MessagingConstants.java` | New `PUSH_NOTIFICATION_RECEIVED = "pushnotificationreceived"` data key for routing. |

## XDM payload on the wire to Edge

```json
{
  "xdm": {
    "eventType": "pushTracking.receive",
    "pushNotificationTracking": {
      "pushProvider": "fcm",
      "pushProviderMessageId": "<FCM messageId>"
    },
    "application": {
      "launches": {
        "value": 0
      }
    },
    "_experience": {
      "customerJourneyManagement": {
        /* merged from _xdm field of AJO payload */
      }
    }
  },
  "meta": {
    "collect": {
      "datasetId": "<experienceEventDatasetId from Configuration shared state>"
    }
  }
}
```

## Types of changes

## Checklist:


code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/MessagingService.java
```diff
@@ -35,15 +39,26 @@ public class MessagingService extends FirebaseMessagingService {
...
+    // key name written by ConfigurationExtension.configureWithAppID).
+    private static final String CONFIG_DATASTORE = "AdobeMobile_ConfigState";
+    private static final String CONFIG_KEY_APP_ID = "config.appID";
+    private static final long SELF_INIT_TIMEOUT_SECONDS = 5;
+    private static volatile boolean selfInitTried = false;
+
     @Override
     public void onNewToken(final @NonNull String token) {
         super.onNewToken(token);
+        android.util.Log.d("Akhil", "New Token is  " + token);
```
ishwetansh: can we update this log message.

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/MessagingService.java
```diff
@@ -35,15 +39,26 @@ public class MessagingService extends FirebaseMessagingService {
...
+
         MobileCore.setPushIdentifier(token);
     }
 
     @Override
     public void onMessageReceived(final @NonNull RemoteMessage remoteMessage) {
         super.onMessageReceived(remoteMessage);
+        android.util.Log.d(
```
ishwetansh: update log.

code/messaging/src/phone/java/com/adobe/marketing/mobile/Messaging.java
```diff
@@ -68,6 +71,25 @@ public final class Messaging {
     private static final String REFRESH_MESSAGES_EVENT = "refreshmessages";
     private static final String RESPONSE_ERROR = "responseerror";
     private static final String SCOPE = "scope";
+    private static final String PUSH_NOTIFICATION_RECEIVED = "pushnotificationreceived";
```
ishwetansh, line 74: PUSH_NOTIFICATION_RECEIVED constant has a duplicate entry

code/messaging/src/phone/java/com/adobe/marketing/mobile/Messaging.java
```diff
@@ -136,9 +158,67 @@ public static boolean addPushTrackingDetails(
...
+     * Dispatches a push notification received event to track that a push notification was
+     * displayed. This is called internally by both the SDK-handled path ({@link
+     * MessagingService#handleRemoteMessage}) and the client-handled path ({@link
+     * #addPushTrackingDetails}).
+     *
+     * @param messageId {@link String} the Firebase message ID from {@code
+     *     RemoteMessage#getMessageId()}
+     * @param data {@link Map} the data payload from {@code RemoteMessage#getData()}
+     */
+    public static void handlePushReceived(
```
ishwetansh: If this API is called internally, then any specific reason to make it public ?

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/MessagingService.java
```diff
@@ -66,19 +94,145 @@ public static boolean handleRemoteMessage(
...
+     * redundant registrations are short-circuited by the Core EventHub with no side effects.
+     *
+     * @param context the {@link Context} from FCM's {@code onMessageReceived}. Must have a non-null
+     *     {@link Application} as its application context.
+     */
+    private static synchronized void selfInit(final @NonNull Context context) {
+        if (selfInitTried) {
+            return;
+        }
+        selfInitTried = true;
```
ishwetansh, line 112: latch only when we're actually calling initialize

ishwetansh: due to some pre-condition code could exit from the if block

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/MessagingService.java
```diff
@@ -66,19 +94,145 @@ public static boolean handleRemoteMessage(
...
-        final Event pushNotificationReceivedEvent =
-                new Event.Builder(
-                                "Push Notification Displayed",
-                                EventType.MESSAGING,
-                                EventSource.RESPONSE_CONTENT)
-                        .setEventData(notificationData)
-                        .build();
-        MobileCore.dispatchEvent(pushNotificationReceivedEvent);
+        // track push notification received
+        Messaging.handlePushReceived(remoteMessage.getMessageId(), remoteMessage.getData());
```
ishwetansh: getMessageId() can return null - 

it would be a better idea to have a null check in place
`  final String messageId = remoteMessage.getMessageId();
  if (StringUtils.isNullOrEmpty(messageId)) {
      Log.warning(MessagingPushConstants.LOG_TAG, SELF_TAG,
              "handleRemoteMessage: messageId is null, skipping.");
      return false;
  }`

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/MessagingExtension.java
```diff
@@ -43,6 +43,26 @@
...
+     * #onUnregistered()}. Volatile because it is read from threads other than the EventHub executor
+     * (specifically, the FCM service thread).
+     */
+    private static volatile boolean registered = false;
+
+    /**
+     * @return {@code true} if {@link MessagingExtension} is currently registered with Core's
+     *     EventHub in this process, {@code false} otherwise.
+     */
+    public static boolean isRegistered() {
```
ishwetansh: does it needs to be public ?

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/MessagingService.java
```diff
@@ -66,19 +94,145 @@ public static boolean handleRemoteMessage(
...
+        }
+    }
+
+    /**
+     * Reads the {@code appId} previously written to {@value #CONFIG_DATASTORE} / {@value
+     * #CONFIG_KEY_APP_ID} by Core's {@code ConfigurationExtension.configureWithAppID}.
+     *
+     * @return the persisted appId, or {@code null} if not present or unreadable.
+     */
+    @androidx.annotation.Nullable private static String readCachedAppId() {
```
ishwetansh: Nit - would be a better idea to put readCachedAppId in the next line after the annotation
` 
@androidx.annotation.Nullable

private static String readCachedAppId() {
`

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/MessagingService.java
```diff
@@ -66,19 +94,145 @@ public static boolean handleRemoteMessage(
...
+                "Self-init: cached appId found, calling MobileCore.initialize.");
+
+        // Disable automatic lifecycle tracking — we're on the FCM service thread, not in an
+        // Activity context. Lifecycle events should be driven by the host app's foreground init.
+        final InitOptions options = InitOptions.configureWithAppID(cachedAppId);
+        options.setLifecycleAutomaticTrackingEnabled(false);
+
+        final CountDownLatch latch = new CountDownLatch(1);
+        try {
+            MobileCore.initialize(application, options, ignored -> latch.countDown());
```
ishwetansh: Curious about the condition here, Do we need to block thread for 5 seconds, instead can't we put the code ` Messaging.handlePushReceived(messageId, data);` inside initialize callback?

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/MessagingService.java
```diff
@@ -66,19 +94,145 @@ public static boolean handleRemoteMessage(
...
+        } catch (final RuntimeException e) {
```
sagar-sharma-adobe: Can we remove try catch from here if there is no threat of exception.

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/MessagingExtension.java
```diff
@@ -396,6 +419,32 @@ void processEvent(final Event eventToProcess) {
...
+                    getSharedState(
+                            MessagingConstants.SharedState.Configuration.EXTENSION_NAME,
+                            eventToProcess);
+            final String receivedDatasetId =
+                    DataReader.optString(
+                            receivedConfigSharedState,
+                            MessagingConstants.SharedState.Configuration
+                                    .EXPERIENCE_EVENT_DATASET_ID,
+                            "");
+            if (StringUtils.isNullOrEmpty(receivedDatasetId)) {
```
navratan-soni, line 415: Is it really necessary to check dataset ID

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/MessagingService.java
akhiljain1907: this was for testing purpose. I have updated it to SDK pattern and can be removed before final merge :+1:

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/MessagingService.java
akhiljain1907: updated to SDK used pattern

code/messaging/src/phone/java/com/adobe/marketing/mobile/Messaging.java
akhiljain1907: This mirrors MessagingConstants.EventDataKeys.Messaging.PUSH_NOTIFICATION_RECEIVED. Cross-package access isn't possible (the inner class is package-private), so this matches the existing pattern used by the other 16 cross-file constants in Messaging.java (e.g., TRACK_INFO_KEY_MESSAGE_ID, TRACK_INFO_KEY_ADOBE_XDM, EVENT_SOURCE_NOTIFICATION).

code/messaging/src/phone/java/com/adobe/marketing/mobile/Messaging.java
akhiljain1907: As discussed with @ashumadobe, we should have a public api for handle tracking and calling it automatically in addPushTrackingDetails can be discussed with product. We will setup a call with product to take decision on this.

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/MessagingService.java
akhiljain1907: removed countdown latch logic

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/MessagingExtension.java
akhiljain1907: updated the logic in handlerRemoteMessage and this is now used in only same package class. So made it internal from public. Thanks :blush:

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/MessagingService.java
akhiljain1907: Done :thumbsup:

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/MessagingService.java
akhiljain1907: Initially countdown latch was kept to make sure we have enough time to register our extensions. Removed countdown latch logic and called handlePushReceived from callback. Will note the TAT for initialize on different networks to check the impact, but this definitely is the ideal approach.

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/MessagingService.java
akhiljain1907: Removed

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/MessagingExtension.java
akhiljain1907: Yes — the datasetId is the routing field for the Edge event (meta.collect.datasetId). Without it, the event reaches Edge but can't be routed to cjm_push_tracking_experience_event_dataset, so it would silently fail to ingest. The same check exists for handleTrackingInfo downstream — removing the check here would make the two paths inconsistent. The warning log also helps customers debug Launch property misconfiguration (no messaging extension / no experience event dataset bound).

akhiljain1907:
Hi @ishwetansh, @sagar-sharma-adobe, @navratan-soni many thanks for review. I have resolved your comments and as discussed, we initially kept countdown latch for POC purpose with a 5 second hardcode to ensure our SDK has enough time to register extensions. As calling handlePushReceieved in MobileCore.initialize() api's callback is ideal, i have updated the code to call push tracking event from initialize's callback. We will perform our further testing on this path and see if there is any difference in results. I am pasting these two diagrams which explains before and after design - 

       **Before - (latch based)**
       [FCM service thread]                       [EventHub executor]
       │
       │ onMessageReceived()
       │
       ├─► handleRemoteMessage()
       │       │
       │       │ isAJONotification? yes
       │       │ isRegistered? no
       │       │
       │       ├─► selfInit()
       │       │       │
       │       │       │ setApplication
       │       │       │ readCachedAppId → "ENV-FILE-ID"
       │       │       │ new CountDownLatch(1)
       │       │       │
       │       │       ├─► MobileCore.initialize(callback) ─────► scope.launch
       │       │       │                                                │
       │       │       │ ╔══════════════════════════════╗               │ discover extensions
       │       │       │ ║  latch.await(5s)             ║               │ register Configuration
       │       │       │ ║  ◀── BLOCKED 250-900 ms ──▶  ║               │ register Edge
       │       │       │ ║  (worst case 5 s timeout)    ║               │ register Identity
       │       │       │ ╚══════════════════════════════╝               │ register Messaging
       │       │       │                                                │
       │       │       │ ◀──────── latch.countDown() ◀─── initialize completion callback
       │       │       │
       │       │   selfInit returns
       │       │
       │       │ build notification
       │       │ NotificationManager.notify(...)   ← USER SEES IT 5s LATE
       │       │
       │       │ Messaging.handlePushReceived(messageId, data)
       │       │     dispatchEvent(MESSAGING/REQUEST_CONTENT) ──────────► event in queue
       │       ◀── return true ──                                        │
       │                                                                 │ MessagingExtension.processEvent
       │ onMessageReceived returns                                       │ → handleTrackingInfo
       │ FCM wakelock released                                           │ → dispatch EDGE event
       │ (process at risk of being killed)                               │     ───────────────► [Edge executor]
                                                                                                       │
                                                                                                       │ HTTP POST
                                                                                                       │ ...
       
       
       
       **After - (Callback based)**
       [FCM service thread]                       [EventHub executor]
       │
       │ onMessageReceived()
       │
       ├─► handleRemoteMessage()
       │       │
       │       │ isAJONotification? yes
       │       │ messageId null? no  ← (new null guard, ~1 ms)
       │       │
       │       │ build notification
       │       │ NotificationManager.notify(...)   ← USER SEES IT IMMEDIATELY
       │       │
       │       │ isRegistered? no
       │       │
       │       ├─► selfInit(context, runnable)
       │       │       │
       │       │       │ setApplication
       │       │       │ readCachedAppId → "ENV-FILE-ID"
       │       │       │
       │       │       ├─► MobileCore.initialize(app, opts, runnable-wrapper) ──► scope.launch
       │       │       │                                                                │
       │       │       ◀── selfInit returns immediately (~5 ms total) ──                │
       │       │                                                                        │ register Configuration
       │       ◀── handleRemoteMessage returns true ──                                  │ register Edge
       │                                                                                │ register Identity
       │ onMessageReceived returns IN ~10 MS                                            │ register Messaging
       │ FCM wakelock released                                                          │
       │ (process stays alive on its own grace window)                                  │
       │                                                                                │ initialize completion callback fires:
       │                                                                                │   runnable.run()
       │                                                                                │     = Messaging.handlePushReceived(messageId, data)
       │                                                                                │         dispatchEvent(MESSAGING/REQUEST_CONTENT)
       │                                                                                │
       │                                                                                │ MessagingExtension.processEvent (same thread)
       │                                                                                │ → handleTrackingInfo
       │                                                                                │ → dispatch EDGE event
       │                                                                                │     ───────────────► [Edge executor]
       │                                                                                                            │
       │                                                                                                            │ HTTP POST
       │                                                                                                            │ ...

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/MessagingService.java
```diff
@@ -58,27 +75,168 @@ public static boolean handleRemoteMessage(
...
+        }
+        Log.debug(
+                MessagingPushConstants.LOG_TAG,
+                SELF_TAG,
+                "Self-init: cached appId found, calling MobileCore.initialize.");
+
+        // Disable automatic lifecycle tracking — we're on the FCM service thread, not in an
+        // Activity context. Lifecycle events should be driven by the host app's foreground init.
+        final InitOptions options = InitOptions.configureWithAppID(cachedAppId);
+        options.setLifecycleAutomaticTrackingEnabled(false);
```
sagar-sharma-adobe: We are disabling lifecycle automatic tracking for application here. Is this safe? Since only one initialize can heppen this may lead yo lifecycle tracking being disabled for the app.

akhiljain1907: Good catch @sagar-sharma-adobe — Core's once-per-process initialize contract, if self-init ran first (killed-state cold-start), our disabled-lifecycle setting would stick for the rest of the process and the host app's later MobileCore.initialize(app, appId) call couldn't restore it.

Fix: switched to MobileCore.initialize(app, appId, callback) — the simpler overload that gives us platform-default options (lifecycle tracking ON, matching what the simple host API gives). Updated the Javadoc to call out the constraint: customers needing non-default options must initialize from Application.onCreate, since cold-start pushes can preempt later init calls. This is consistent with the existing recommendation to initialize in Application.onCreate anyway.

sagar-sharma-adobe (approved):

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/MessagingExtension.java
```diff
@@ -43,6 +43,26 @@
...
+     * #onUnregistered()}. Volatile because it is read from threads other than the EventHub executor
+     * (specifically, the FCM service thread).
+     */
+    private static volatile boolean registered = false;
+
+    /**
+     * @return {@code true} if {@link MessagingExtension} is currently registered with Core's
+     *     EventHub in this process, {@code false} otherwise.
+     */
+    static boolean isRegistered() {
```
sagar-sharma-adobe: Let's validate if we need this flag. Or can we remove it?

code/messaging/src/phone/java/com/adobe/marketing/mobile/Messaging.java
```diff
@@ -139,6 +161,72 @@ public static boolean addPushTrackingDetails(
...
+     *
+     * <p>A bounded in-memory dedup cache (cap 10, keyed on {@code messageId}) ensures the receive
+     * event is dispatched exactly once per notification per process, even if this method is called
+     * multiple times for the same message.
+     *
+     * @param messageId {@link String} the Firebase message ID from {@code
+     *     RemoteMessage#getMessageId()}
+     * @param data {@link Map} the data payload from {@code RemoteMessage#getData()}
+     */
+    public static void handlePushReceived(
```
sagar-sharma-adobe: Can we instead work with single param called remoteMessage?
Move messageId extraction code to this function only so that the app doesn't have to do it.

sagar-sharma-adobe (changes requested):

code/messaging/src/phone/java/com/adobe/marketing/mobile/Messaging.java
```diff
@@ -68,6 +72,25 @@ public final class Messaging {
...
+    private static final String PUSH_NOTIFICATION_RECEIVED = "pushnotificationreceived";
+    private static final String EVENT_TYPE_PUSH_TRACKING_RECEIVED = "pushTracking.receive";
+
+    // Bounded in-memory cache to deduplicate push receive events.
+    // Guards against multiple handlePushReceived calls for the same messageId within a process
+    // lifetime — e.g. FCM at-least-once burst delivery after a FLAG_STOPPED window clears.
+    private static final int MAX_DEDUP_CACHE_SIZE = 10;
+
+    @SuppressWarnings("serial")
+    private static final Set<String> recentlyTrackedMessageIds =
```
navratan-soni: Remove de dup logic, it is not needed

code/messaging/src/phone/java/com/adobe/marketing/mobile/Messaging.java
```diff
@@ -139,6 +162,73 @@ public static boolean addPushTrackingDetails(
...
+     * </ul>
+     *
+     * <p>A bounded in-memory dedup cache (cap 10, keyed on {@code messageId}) ensures the receive
+     * event is dispatched exactly once per notification per process, even if this method is called
+     * multiple times for the same message.
+     *
+     * @param remoteMessage {@link RemoteMessage} the Firebase remote message received in {@code
+     *     FirebaseMessagingService#onMessageReceived}
+     */
+    public static void handlePushReceived(@NonNull final RemoteMessage remoteMessage) {
```
navratan-soni: Can we change this to trackPushReceive, as this is only used for tracking

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/MessagingService.java
```diff
@@ -58,27 +76,120 @@ public static boolean handleRemoteMessage(
...
-                                EventType.MESSAGING,
-                                EventSource.RESPONSE_CONTENT)
-                        .setEventData(notificationData)
-                        .build();
-        MobileCore.dispatchEvent(pushNotificationReceivedEvent);
+        // Bootstrap the SDK if this is a cold-start push, then record delivery. selfInit is a
+        // no-op after the first call (guarded by selfInitTried) and MobileCore fires its
+        // initialize callback even when already initialized, so this works correctly for both
+        // cold-start and warm-start pushes.
+        selfInit(context, () -> Messaging.handlePushReceived(remoteMessage));
```
navratan-soni: I think we can move slefInit inside handlePushreceived, 
in case custom service implemenation, when this is being called (as public api) it will ensure that we are able to track

sagar-sharma-adobe (approved):

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/MessagingService.java
akhiljain1907: As discussed we will keep self init only in handleRemoteMessage and for trackPushReceived we will document that the customers should call it after SDK is initialized.

code/messaging/src/phone/java/com/adobe/marketing/mobile/Messaging.java
akhiljain1907: Done

code/messaging/src/phone/java/com/adobe/marketing/mobile/Messaging.java
akhiljain1907: Done

PR #449, opened 2026-08-03
Add offline content card availability with disk persistence

akhiljain1907:
Persist content card propositions to disk so they are available offline at boot before the initial network fetch. Track card origin (NETWORK vs DISK) in memory to enrich interaction XDM with servedFromPersistentCache. Handle non-recoverable Edge errors by preserving last-known-good state. Clear persisted and In Memory CC state on identity reset.

## Types of changes

## Checklist:


code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/EdgePersonalizationResponseHandler.java
```diff
@@ -85,6 +88,12 @@ class EdgePersonalizationResponseHandler {
     // holds content cards that the user has qualified for
     private Map<Surface, List<Proposition>> contentCardsBySurface = new HashMap<>();
 
+    // tracks the origin (NETWORK or DISK) of content card rules per proposition id
+    private final Map<String, CardOrigin> contentCardOriginByProposition = new HashMap<>();
```
sagar-sharma-adobe: Card origin needs to be saved with the cached proposition itself. Let's keep implementation on both android and iOS semantically same.

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/EdgePersonalizationResponseHandler.java
```diff
@@ -706,20 +760,69 @@ private void applyPropositionChangeForEventId(final String eventId) {
...
 
+        // disk-first write strategy: persist content card and inbox propositions to disk
+        // before updating in-memory state, ensuring crash consistency
+        if (offlineAvailable) {
+            if (!MapUtils.isNullOrEmpty(parsedPropositions.contentCardPropositionsToPersist)) {
+                messagingCacheUtilities.cacheContentCardPropositions(
+                        parsedPropositions.contentCardPropositionsToPersist, surfacesToRemove);
+            }
+            if (!MapUtils.isNullOrEmpty(parsedPropositions.inboxPropositionsToPersist)) {
+                messagingCacheUtilities.cacheInboxPropositions(
```
sagar-sharma-adobe: Are we persisting inbox as well? iOS is not currently persisting inbox. Let's make sure both android and iOS are on the same page.

code/messaging/src/phone/java/com/adobe/marketing/mobile/Messaging.java
```diff
@@ -506,6 +510,151 @@ public void call(final Event event) {
...
+     *     previously fetched propositions
+     * @param usePersistedContentCards If {@code true}, retrieves propositions from the persisted
+     *     disk cache; if {@code false}, behaves identically to {@link
+     *     #getPropositionsForSurfaces(List, AdobeCallback)}
+     * @param callback A {@link AdobeCallback} which will be invoked with a {@link Map<Surface,
+     *     List<Proposition>>} containing previously fetched content card or code based content
+     */
+    public static void getPropositionsForSurfaces(
+            @NonNull final List<Surface> surfaces,
+            final boolean usePersistedContentCards,
```
sagar-sharma-adobe: As per latest discussions this flag was no longer needed. Please remove this. Only global flag is going to be utilized for gating offline feature.

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/EdgePersonalizationResponseHandler.java
```diff
@@ -1193,6 +1296,400 @@ void createInAppMessage(final PropositionItem propositionItem) {
...
+    /**
+     * Hydrates content card rules engine from persisted disk cache. Called at boot time before the
+     * initial network fetch to provide offline content card availability.
+     *
+     * <p>Also loads event-history rules (disqualify/dismiss) into the main rules engine so that
+     * dismiss/disqualify operations work for disk-loaded content cards without requiring a network
+     * response first.
+     */
+    void hydrateContentCardRulesEngineFromDisk() {
+        if (!isContentCardOfflineAvailable()) {
```
sagar-sharma-adobe: The hydration can be ungated. iOS also has ungated hydration.

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/EdgePersonalizationResponseHandler.java
```diff
@@ -706,20 +760,69 @@ private void applyPropositionChangeForEventId(final String eventId) {
...
+        final List<Surface> ccRequestedSurfaces;
+        if (offlineAvailable) {
+            ccRequestedSurfaces = new ArrayList<>(requestedSurfaces);
+            final Map<Surface, List<LaunchRule>> ccNewRules =
+                    surfaceRulesBySchemaType.get(SchemaType.CONTENT_CARD);
+            final Iterator<Surface> it = ccRequestedSurfaces.iterator();
+            while (it.hasNext()) {
+                final Surface surface = it.next();
+                final boolean hasNewRules = ccNewRules != null && ccNewRules.containsKey(surface);
+                if (!hasNewRules && hasDiskOriginContentCards(surface)) {
```
sagar-sharma-adobe: This exclusion breaks legitimate eviction. If a surface was requested and the server successfully returns no content card for it (campaign ended), removing it from ccRequestedSurfaces means removeOrReplaceContentCards never evicts the disk card — so it lingers in memory for the session and is still tracked as servedFromPersistentCache=true.

It's also unnecessary for the case it targets: a disk card on a surface that wasn't requested is already safe, because removeOrReplaceContentCards only iterates the requested surfaces.

Suggest dropping this exclusion and matching iOS's rule: a successful response is authoritative for every requested surface (replace if returned, evict if empty); surfaces that weren't requested are left untouched. That preserves disk cards for unrelated requests without blocking real eviction.

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/EdgePersonalizationResponseHandler.java
```diff
@@ -159,6 +182,70 @@ class EdgePersonalizationResponseHandler {
...
+
+    /**
+     * Determines whether the device currently has internet connectivity, used to short-circuit
+     * user-triggered proposition fetches when offline. Fails open (returns {@code true}) when
+     * connectivity cannot be determined, so a fetch is never wrongly suppressed.
+     *
+     * @return {@code true} if internet is available or connectivity is indeterminate
+     */
+    boolean isInternetAvailable() {
+        if (internetAvailableOverrideForTesting != null) {
```
sagar-sharma-adobe: Is this flag needed here in the function implemetation?

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/EdgePersonalizationResponseHandler.java
```diff
@@ -1193,6 +1348,267 @@ void createInAppMessage(final PropositionItem propositionItem) {
...
+        // card
+        // served for them reports servedFromPersistentCache = true until a live network response
+        // refreshes the surface this session.
+    }
+
+    /**
+     * Hydrates all persisted content cards from disk. Called during boot-time setup before the
+     * initial network fetch.
+     */
+    void hydrateAllPersistedContentCards() {
```
sagar-sharma-adobe: This seems to be a wrapper only without any particular implementation

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/ParsedPropositions.java
```diff
@@ -138,12 +149,16 @@ public class ParsedPropositions {
                         break;
                     case JSON_CONTENT:
                     case HTML_CONTENT:
-                    case INBOX:
                     case DEFAULT_CONTENT:
                         propositionsToCache =
                                 MessagingUtils.updatePropositionMapForSurface(
                                         surface, proposition, propositionsToCache);
                         break;
+                    case INBOX:
```
sagar-sharma-adobe: There seems to be no difference in the code change. Is there a purpose for this change?

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/EdgePersonalizationResponseHandler.java
```diff
@@ -554,6 +641,19 @@ void retrieveInMemoryPropositions(final List<Surface> surfaces, final Event even
             return;
         }
 
+        // When offline availability is disabled, drop content cards for surfaces that were not
+        // refreshed from the network this session (i.e. loaded from disk at boot), and evict the
+        // stale disk cache. This handles disabling the flag at runtime. Only content card state is
+        // affected — inbox and code-based propositions live in a separate store and are always
+        // served for the requested surfaces.
+        if (!isContentCardOfflineAvailable()) {
+            contentCardsBySurface.keySet().retainAll(networkRefreshedSurfaces);
```
sagar-sharma-adobe: I don't think we should be modifying this map.

akhiljain1907: Done, instead of directly modifying the contentCardsBySurface map we now create a copy and then filter from it

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/EdgePersonalizationResponseHandler.java
akhiljain1907: Non this flag is defined like - 
  @VisibleForTesting static Boolean internetAvailableOverrideForTesting = null;
  
  And is only used for tests

code/messaging/src/main/java/com/adobe/marketing/mobile/messaging/EdgePersonalizationResponseHandler.java
akhiljain1907: Yes this was just a wrapper to hydrate Inbox and other inbound channels as well, since now we are handling only content card i have removed this wrapper

sagar-sharma-adobe (approved):

sagar-sharma-adobe (approved):

